PR #46, opened 2018-05-25
Move getting started content and make it more generic

dedemorton:
Notes:

* This is a rewrite of the getting started content in the Beats Platform Reference. Once we finalize this content and add it to the stack docs, I'll remove it from the Platform Reference in versions 6.3 and later. I currently have the content in a stand-alone book to make it easier to build/review.

* A lot of the content has changed. Please review the full document and respond to my questions.

* To build the content, pull down the files and from the `docs` repo, run:

`build_docs.pl --doc ../stack-docs/docs/en/getting-started/index.asciidoc --chunk 1 -open`


dedemorton:
@lcawl @debadair Not sure who else wants to review this.

@tsg Can you suggest the folks on the dev team who should review this?

docs/en/getting-started/index.asciidoc
```diff
@@ -0,0 +1,19 @@
+[[elastic-stack-getting-started]]
+= Getting started with the Elastic Stack
+
+:ls: Logstash
```
dedemorton: These attributes need to be moved to the shared attributes list.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
+[[get-started-elastic-stack]]
+== Getting started with the {stack}
+
+Looking for an {stack} ("ELK tutorial") that shows how to set up the {stack}? In
```
lcawl: Just a small comment, but I think "tutorial" should be outside these brackets.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+
+*win:*
+
+["source","sh",subs="attributes,callouts"]
+----------------------------------------------------------------------
+bin\logstash.bat -f logstash.conf
+----------------------------------------------------------------------
+
+NOTE: The default configuration for {beats} and {ls} uses plain TCP. For
+encryption you must explicitly enable SSL when you configure {beats} and {ls}.
```
lcawl: Is there somewhere they can find more information about that?  Maybe https://www.elastic.co/guide/en/elastic-stack-overview/master/encrypting-communications.html ?

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+{ls} reads from the {beats} input and indexes events into {es}. You haven't
+defined a filter section yet, so {ls} simply forwards events to {es} without
+additional processing. Next, you learn how to define the filter stage.
+
+[float]
+[[logstash-filter]]
+==== Define a filter to extract data from a field
+
+The system metrics collected by {metricbeat} include a field called `cmdline`
+that contains the full command-line argument used to start the process. For
```
lcawl: I don't know Logstash or Metricbeat very well at all, so it wasn't clear to me what "the process" was here.  I'm presuming that there are a specific set of processes that Metricbeat collects info about, but it might be nice to be clearer if possible.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+      }
+    }
+  }
+}
+----
+
+
+Learning grok is beyond the scope of this tutorial, but if you want to learn
+more, see the {logstash-ref}/plugins-filters-grok.html[Grok filter plugin]
+documentation.
```
lcawl: Maybe this should be moved up sooner?  I hit the first occurrence of "grok filter" in the paragraph above the example and went down a rabbit hole trying to figure out whether I was supposed to know what a grok filter was.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+IMPORTANT: Implementing security is a critical step in setting up the {stack}.
+To get up and running quickly with a sample installation, you skip those steps
+right now, but keep in mind that security is a critical component of your
+deployment.
+
+// REVIEWERS: I haven't discussed security because it really complicates the learning
+// experience. I think there's a value to teaching users how to set up a basic
+// installation (and get the various components talking to each other) before
+// adding security.
+
```
lcawl: Agreed!

docs/en/getting-started/get-started-stack.asciidoc
dedemorton: Yah I should beef up the intro with that info actually!

docs/en/getting-started/get-started-stack.asciidoc
andrewkroh: I'm good with removing the note if there's a security step.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+  hosts: ["localhost:5044"]
+----
+
+
+Save the file, then restart {metricbeat} to apply the changes.
+
+
+[float]
+[[logstash-setup]]
+==== Configure {ls} to read from the {beats} input
```
yaauie: I would suggest retitling this section `Configure {ls} to listen for {beats} input`.

When started with this pipeline, Logstash will passively listen on the configured port for new connections, but the "read from" phrasing makes it sound like Logstash is actively finding the beats and pulling data from them, which could lead to misconceptions and confusion about how the two work together.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+
+{ls} provides {logstash-ref}/input-plugins.html[input plugins] for reading from
+a variety of inputs. In this tutorial, you create a {ls} pipeline configuration
+that reads from the {beats} input and sends events to the {es} output.
```
yaauie:
~~~ diff
-that reads from the {beats} input and sends events to the {es} output.
+that listens for {beats} input and sends the received events to the {es} output.
~~~

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+
+To configure {ls}:
+
+. Create a new {ls} pipeline configuration file called `logstash.conf` that
```
yaauie: I would suggest using a more descriptive name for the pipeline configuration (perhaps: `demo-metrics-pipeline.conf`, which enables a user who is looking at that directory to have a cursory understanding of the purpose of the pipeline configuration without digging into its contents).

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+----------------------------------------------------------------------
+
+*rpm:*
+
+["source","sh",subs="attributes,callouts"]
+----------------------------------------------------------------------
+sudo service logstash start
+----------------------------------------------------------------------
+
+*mac:*
```
yaauie: this command also works for manually-installed Logstash on Linux.

docs/en/getting-started/get-started-stack.asciidoc
yaauie: "plain TCP" is an implementation detail that I don't think is helpful here; how about:

~~~
NOTE: the default configuration for {beats} and {ls} does not use encryption and does not
provide secure communication; before sending sensitive data across the network, make sure
you enable {stack-ref}/encrypting-communications.html[encrypted communications]
between {beats} and {ls}.
~~~

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+
+[source,ruby]
+----
+filter {
+  if [system][process] {
+    if [system][process][cmdline] {
+      grok {
+        match => { "[system][process][cmdline]" => ["%{CMD:[system][process][short_cmdline]}"] }
+        pattern_definitions => {
+          "CMD" => "(?:/[A-Za-z0-9$.+!*'(){},~:;=@#%&_\-]*)+\s*-*[A-Za-z0-9$.+!*'(){},~:;=@#%&_\-]*"
```
yaauie: This pattern is one heck of an introduction to GROK and regular expressions; can we start with something simpler?

What about capturing the just the cmd's path, which would allow us to use a pre-defined `PATH` pattern? That would just be:

~~~
filter {
  if [system][process] {
    if [system][process][cmdline] {
      grok {
        match => {
          "[system][process][cmdline]" => "^%{PATH:[system][process][cmdline_path]}"
        }
        remove_field => "[system][process][cmdline]"
      }
    }
  }
}
~~~

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+-isForBrowser -boolPrefs 36:1|299:0| -stringPrefs 285:38;{b77ae304-9f53-a248-8bd4-a243dbf2cab1}| -schedulerPrefs
+0001,2 -greomni /Applications/Firefox.app/Contents/Resources/omni.ja -appomni
+/Applications/Firefox.app/Contents/Resources/browser/omni.ja -appdir
+/Applications/Firefox.app/Contents/Resources/browser -profile
+/Users/dedemorton/Library/Application Support/Firefox/Profiles/mftvzeod.default-1468353066634
+99468 gecko-crash-server-pipe.99468 org.mozilla.machname.1911848630 tab"
+----
+
+
+You might decide that you don't want to send the whole command-line argument to
```
yaauie: Alternate:

~~~
You might decide that you don't want to send the whole command-line argument to
{es}, and that sending the command's _path_ might be sufficient. One way to do that
would be by using a Grok filter to describe the pattern to extract, and to remove
the original field from the event to ensure it is not indexed. For example:
~~~

docs/en/getting-started/get-started-stack.asciidoc
yaauie: Grok can be exceptionally powerful, but it's a super deep rabbit hole to go down before developing solid footing on what Filters bring to the table. Maybe we should start with a different filter?

In the Logstash community forums, we tend to steer people _away_ from Grok until they've exhausted other avenues for extracting data (e.g., often-overlooked tools like the Dissect filter can be significantly simpler to implement and maintain). And, because Grok is a pattern language, it is _necessary_ to start with a corpus of messages; starting with a single message is generally a recipe for [disfluency](https://en.wikipedia.org/wiki/Garden_path_sentence).

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+output {
+  elasticsearch {
+    hosts => "localhost:9200"
+    manage_template => false
+    index => "%{[@metadata][beat]}-%{[@metadata][version]}-%{+YYYY.MM.dd}" <1>
+  }
+}
+----
++
+This configuration allows you to route events to {ls}, where you have full
```
yaauie: The phrasing here leads me to believe that this configuration is the _source_ of the events,
but in reality it is the _recipient_. Perhaps something like:


~~~
When {ls} is started with this pipeline configuration, you will be able to route events
from {beats} to {elasticsearch} through {ls}, where you have full access to the {ls}
capabilities for collecting, enriching, and transforming data.
~~~

docs/en/getting-started/get-started-stack.asciidoc
dedemorton: I'm removing the note from this section, but working some of Ry's language into the note that appears earlier because it's important to call out that communications are not secure until you explicitly turn on security.

docs/en/getting-started/get-started-stack.asciidoc
dedemorton: Make sense! Ironically, that's where I started, but then went down a rabbit hole...I thought I wanted to capture the first argument, but backed away from that because it got weird quickly.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -778,37 +770,35 @@ example:
...
-You might decide that you don't want to send the whole command-line argument to
-{es}. You can write a grok filter that matches the first part of the command
-and saves it to a new field. You can also remove the original field, `cmdline`,
-from the event. For example:
-
-// REVIEWERS: I took a stab at this, but I'm not very good at grok or
-// regex. In fact, I sort of hate it...a lot. Can someone review and suggest
-// something better based on the format you would expect in this field?
+Rather than sending the whole command-line argument to {es}, you might want to
+send just the command's path. One way to do that is by using a Grok filter.
```
dedemorton: I don't like that this is wordier, but I want to make it clear that the grok filter is just one of the ways.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+ifeval::["{release-state}"!="unreleased"]
+
+. Download the {metricbeat} Windows zip file from the
+https://www.elastic.co/downloads/beats/metricbeat[{metricbeat} download] page.
+
+. Extract the contents of the zip file into `C:\Program Files`.
+
+. Rename the ++metricbeat-{version}-windows++ directory to ++{metricbeat}++.
+
+. Open a PowerShell prompt as an Administrator (right-click the PowerShell icon
```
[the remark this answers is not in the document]
dedemorton: @yaauie I'm not sure how to address this comment. Let me know if you have specific suggestions.

docs/en/getting-started/get-started-stack.asciidoc
dedemorton: For now, I've updated the tutorial to reflect your suggestions for the current example. We can always merge and improve later, if necessary.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -0,0 +1,871 @@
...
+https://www.elastic.co/products/elasticsearch[{es}] is a real-time,
+distributed storage, search, and analytics engine. It can be used for many
+purposes, but one context where it excels is indexing streams of semi-structured
+data, such as logs or decoded network packets.
+
+To download and install {es}, open a terminal window and use the commands that
+work with your system (<<deb, deb>> for Debian/Ubuntu, <<rpm, rpm>> for
+Redhat/Centos/Fedora, <<mac, mac>> for OS X, and <<win, win>> for Windows):
+
+// REVIEWERS: Should we add Linux to this list? If so, how do the commands differ
```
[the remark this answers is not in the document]
dedemorton: @yaauie I've updated the instructions to show the tar.gz instead of the .zip, but I think we need to avoid putting too much installation info in this tutorial. We want to keep the installation steps simple and quick...and show the installation method that's the most common for the platforms  we expect users to try this out on. User should look at the installation docs for info about the available packages and installing in a production environment. Sound fair?

dedemorton:
@lcawl @yaauie Can you confirm the changes?

Also, Ry, can you let me know if you want to help with a better example? I'd like to get the merged before the end of the week, if possible. We can always do follow-up PRs because what we have now is quite a bit more useful than what we have live in the Beats Platform Reference.  Thanks!

dedemorton:
@lcawl I've moved the GS content so that it lives in the stack overview. Can you give this a sanity check.

lcawl:
Works for me, thanks!

yaauie:
My apologies for the delay; I was on vacation the last half of last week and missed the notification :/

LGTM. Examples greatly improved by simplifying the grok stuff.

PR #60, opened 2018-06-12
[Docs] Backport to 6.3: move getting started content and make it more generic 

dedemorton:


dedemorton:
I think it does directly follow the intro? Anyhow, I will merge and we can revisit after it's published.

PR #63, opened 2018-06-19
Use .\metricbeat.exe when executing from the CWD

andrewkroh:


dedemorton:
This needs to be backported to 6.3.

PR #65, opened 2018-06-19
Fix broken link

dedemorton:
The content from installing-beats.html is now available in getting-started.html (it's basically just a set of links to the GS guides for all the Beats.


dedemorton:
Need to make this in 6.3 branch, too.

PR #66, opened 2018-06-20
[DOCS]Add merge, reindex, shrink, and split to glossary. Closes #31390

Sue-Gallagher:


docs/en/glossary/glossary.asciidoc
```diff
@@ -614,6 +629,14 @@ some other application.
 endif::logstash-terms[]
 ifdef::elasticsearch-terms[]
 
+[[glossary-shrink]] shrink ::
+
+To reduce the amount of shards in an index. See the {es-ref}/current/indices-shrink-index.html[shrink index API].
++
```
lcawl: I'm not familiar with where/if es-ref is defined. Typically use {ref} instead from https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc

docs/en/glossary/glossary.asciidoc
```diff
@@ -624,6 +647,13 @@ from an ID.
 +
 //Source: Elasticsearch
 endif::elasticsearch-terms[]
+ifdef::elasticsearch-terms[]
+
+[[glossary-split]] split ::
+
+To grow the amount of shards in an index. See the {es-ref}/current/indices-split-index.html[split index API].
+//Source: Elasticsearch
```
lcawl: Ditto

PR #68, opened 2018-06-25
[DOCS] Adds tutorial for getting started with security

lcawl:
The Stack Overview contains a tutorial for installing Elasticsearch, Kibana, Logstash, and Metricbeat. 

This PR creates a new tutorial that describes how to set up user IDs and passwords in the native realm in that environment.

It does not cover how to encrypt communications, however that can be added in this (or another) tutorial hereafter.


lcawl:
You can preview this tutorial here: https://doc-review-63bf6.firebaseapp.com/es2/security-getting-started.html

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+https://www.elastic.co/subscriptions.
+
+You can start a 30-day trial to try out all of the Platinum features, including 
+{security}. Click **Start trial**.
+
+IMPORTANT: If your cluster has already activated a trial license for the current 
+major version, you cannot start a new trial. For example, if you have already 
+activated a trial for v6.0, you cannot start a new trial until v7.0.
+
+At the end of the trial period, your license reverts to a Basic license. If you
```
tvernum: Is this correct?
I didn't think we reverted automatically, but I may have missed something.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+At the end of the trial period, your license reverts to a Basic license. If you 
+want to keep using the extra features, you can purchase a subscription. 
+
+[float]
+[[gs-security-enabled]]
+=== Enable security in {es}
+
+When you use the trial license, {security} is disabled by default. You must 
+stop {kib}, stop {es}, then add the `xpack.security.enabled` setting to the 
+`config/elasticsearch.yml` file in your {es} directory. For example:
```
tvernum: This is not really specific to this set of docs, but I've noticed we tend to assume that people reading the docs are using the default zip/tar.gz layout with a `config/` directory inside the elasticsearch dir, and not the deb/rpm layout with the config directory in `/etc/elasticsearch` (or the msi install where you get to pick your location, but at least it defaults to being in the elasticsearch directory).

I don't know what we should do to solve that in general, but maybe here we need to have a note for users of the rpm/deb to check the location of their config directory?
- RPM: https://www.elastic.co/guide/en/elasticsearch/reference/6.3/rpm.html#rpm-configuring
- DEB: https://www.elastic.co/guide/en/elasticsearch/reference/6.3/deb.html#deb-configuring

I sometimes get caught out on the forums, where my references to the `config/` directory leave rpm/deb users very confused.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+
+[role="screenshot"]
+image::security/images/management-builtin-users.jpg["User management screenshot in Kibana"]
+
+In this example, you can see all of the built-in users. 
+
+Click *Create user* to make a new user. For example, create a user for yourself:
+
+[role="screenshot"]
+image::security/images/create-user.jpg["Creating a user in Kibana"]
```
tvernum: Is it worth mentioning that the roles are explained in an upcoming step?
I was surprised that there was no explanation here, until I scrolled down a little.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+
+// TBD: The Beats docs don't mention the need for the delete privilege, whereas 
+// the Logstash docs do. Which should be used in this case?
+
+[role="screenshot"]
+image::security/images/create-writer-role.jpg["Creating another role in Kibana"]
+
+Now go back to the *Management / Security / Users* page and assign these roles 
+to the appropriate users. Assign the `metricbeat_reader` role to your personal 
+user.  Assign the `metricbeat_writer` role to the `metricbeat_internal` user.
```
tvernum: I think it's worth mentioning that the personal user should also have `kibana_user`.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+using the native realm. You learned how to create user IDs and roles that 
+prevent unauthorized access to the {stack}. 
+
+Next, you'll want to try other features that are unlocked by your trial license, 
+such as {ml}. See <<ml-getting-started,Getting started with {ml}>>. 
+
+Later, when you're ready to increase the number of nodes in your cluster or set 
+up an production environment, you'll want to encrypt communications across the 
+{stack}. To learn how, read <<encrypting-communications>>. 
+
```
tvernum: Perhaps mention that the `logstash_system` and `beats_system` user are used for monitoring, and refer to those docs?
We told them to remember those passwords when they ran `elasticsearch-setup-passwords`, but we haven't told them what to do with those users.

docs/en/stack/security/get-started-security.asciidoc
jaymode: I don't think it is correct either; the license expires and elasticsearch may operate in a degraded state if security is enabled. That is stats and health apis are not available.

docs/en/stack/security/get-started-security.asciidoc
jaymode: At one point we used $CONF_DIR in the x-pack docs because of this

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+stop {kib}, stop {es}, then add the `xpack.security.enabled` setting to the 
+`config/elasticsearch.yml` file in your {es} directory. For example:
+
+[source,yaml]
+----
+xpack.security.enabled: true
+----
+
+NOTE: This tutorial involves a single node cluster, but if you had multiple 
+nodes, you would enable {security} on every node in the cluster. If you have a
```
jaymode: maybe `you would enable {security} on every node in the cluster and configure Transport Layer Security (TLS) for 
+internode-communication, which is beyond the scope of this tutorial.`

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+[[gs-realms]]
+=== Configure authentication
+
+Now that you've set up the built-in users, you need to decide how you want to 
+manage all the other users. 
+
+The {stack} _authenticates_ users to ensure that they are valid. The 
+authentication process is handled by _realms_. You can use one or more built-in 
+realms, such as the native, file, LDAP, PKI, Active Directory, or SAML realms. 
+Alternatively, you can create your own custom realms. In this tutorial, we'll
```
jaymode: maybe a link to custom realms?

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+}
+----
+<1> Specify the `metricbeat_internal` user that you created earlier in this tutorial. 
+<2> Specify the password that you chose for this user ID.
+
+<<gs-start-logstash,Start {ls}>>. 
+
+[float]
+[[gs-metricbeat-security]]
+=== Add users in {metricbeat}
```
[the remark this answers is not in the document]
dedemorton: Or better yet, maybe remove this section header because it's misleading because the section doesn't cover how to add Metricbeat users.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+<<gs-start-metricbeat,Start {metricbeat}>>. 
+
+[float]
+[[gs-view-security]]
+=== View system metrics in {kib}
+
+Log in to {kib} with the user ID that has `metricbeat_reader` role (for example, 
+`jdoe`). 
+
+This user should have authority to visualize the system metrics (for example, on
```
dedemorton: I'd change the wording here because readers might wonder whether there's an additional step required. If they've followed the steps in the tutorial, the user will have the authority to visualize system metrics, right?

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+
+Log in to {kib} with the user ID that has `metricbeat_reader` role (for example, 
+`jdoe`). 
+
+This user should have authority to visualize the system metrics (for example, on 
+the *Discover* page or in the http://localhost:5601/app/kibana#/dashboard/Metricbeat-system-overview[{metricbeat} system overview dashboard]). 
+
+[float]
+[[gs-security-nextsteps]]
+==== What's next?
```
dedemorton: Shouldn't this be a level === heading?

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+    }
+  }
+}
+
+output {
+  elasticsearch {
+    hosts => "localhost:9200"
+    manage_template => false
+    index => "%{[@metadata][beat]}-%{[@metadata][version]}-%{+YYYY.MM.dd}"
+    user => metricbeat_internal <1>
```
dedemorton: I'd put double quotes around the username and password: `user => "metricbeat_internal"` and `password => your_password`

The last time I checked, if the password contains dashes, it won't work without the double quotes (see https://github.com/elastic/logstash/issues/9618). I think this is true for user, too, but I haven't tested it.

dedemorton: Also (this comment applies to other sections, too), I wonder if this tutorial should be showing how to use the secrets keystores (in ES and LS), too, so that users don't have to hard code values in their configs. I know this makes the tutorial more complicated, but most users do not want to have passwords revealed in config files, so it's a very common use case.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,311 @@
...
+    password => your_password <2>
+  }
+}
+----
+<1> Specify the `metricbeat_internal` user that you created earlier in this tutorial. 
+<2> Specify the password that you chose for this user ID.
+
+<<gs-start-logstash,Start {ls}>>.
```
dedemorton: Same comment as before. I think it's better to repeat the command for ease-of-use.

dedemorton: Also, you might want to mention that it might take a few minutes for these changes to be applied. I think I had to restart Logstash twice, but maybe just because I am impatient.

docs/en/stack/security/get-started-security.asciidoc
lcawl: Thanks, I hadn't realized it could be in different places. I've added a tip where we talk about the elasticsearch.yml and kibana.yml files.  I've also created https://github.com/elastic/elasticsearch/pull/31635 to clean up path issues in some other security pages.

docs/en/stack/security/get-started-security.asciidoc
lcawl: You're right, I've revamped so we tell them it's coming and then explicitly add the kibana_user role.

docs/en/stack/security/get-started-security.asciidoc
lcawl: You're right, I've revamped this section to clarify that.

docs/en/stack/security/get-started-security.asciidoc
lcawl: Good idea, done!

lcawl:
Thanks for all the feedback!  I've updated the content on the preview server here: https://doc-review-63bf6.firebaseapp.com/es2/security-getting-started.html

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,398 @@
...
+----
+elasticsearch.username: "kibana"
+elasticsearch.password: "your_password"
+----
+
+Specify the password that you set with the `elasticsearch-setup-passwords` 
+command then save your changes to the file. 
+--
+
+. If you prefer not to put your user ID and password in the `kibana.yml` file,
```
dedemorton: 1 and 2 are mutually exclusive so I would not make them separate steps. Noticed this because I performed step 1 and then saw step 2...and thought...no, I want to do that instead!

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,398 @@
...
+on the `metricbeat-*` indices:
+
+[role="screenshot"]
+image::security/images/create-writer-role.jpg["Creating another role in Kibana"]
+
+Now go back to the *Management / Security / Users* page and assign these roles 
+to the appropriate users. Assign the `metricbeat_reader` role to your personal 
+user.  Assign the `metricbeat_writer` role to the `logstash_internal` user.
+
+The list of users should now contain all of the built-in users as well as the
```
dedemorton: I really like that you include a screen capture here. It's not good Kibana lets you change tabs without prompting you to Save. You can lose stuff.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,398 @@
...
+      grok {
+        match => {
+          "[system][process][cmdline]" => "^%{PATH:[system][process][cmdline_path]}"
+        }
+        remove_field => "[system][process][cmdline]"
+      }
+    }
+  }
+}
+
```
dedemorton: If possible, I suggest having a shared file that contains the input and filter parts shown here. I worry that we'll update the example in one topic and forget to update it in another. Part of me wonders if it's even necessary to show the full config.

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -0,0 +1,398 @@
...
+    index => "%{[@metadata][beat]}-%{[@metadata][version]}-%{+YYYY.MM.dd}"
+    user => "logstash_internal" <1>
+    password => "your_password" <2>
+  }
+}
+----
+<1> Specify the `logstash_internal` user that you created earlier in this tutorial. 
+<2> Specify the password that you chose for this user ID.
+
+If you prefer not to put your user ID and password in the configuration file,
```
dedemorton: I really went down a rabbit hole because I copy/pasted your first config example and then decided that I wanted to use the keystore instead. So I modified my config, but forgot to add the $ (even though I know it's  required). I wasted time recreating the keystore (I thought that I might have made a typo...there's no way to know for sure because when you add a key to the keystore, it doesn't let you confirm the value) .

I might have avoided this particular rabbit hole  if I'd gone directly to the keystore config example first and copied that. 

You might want to reformat this content or add something to intro so that careless readers (like me!) will know up front that they can use a keystore. Might be worth even pointing out the dollar sign.  This was frustrating because it was one of those dumb mistakes that your eyes skip over and the error message is not useful.

docs/en/stack/security/get-started-security.asciidoc
lcawl: Good point. I'll omit everything but the elasticsearch section

lcawl:
Thanks for the feedback, @dedemorton !  I've put the config file and keystore options as bulleted options (e.g. https://doc-review-63bf6.firebaseapp.com/es2/security-getting-started.html#gs-kibana-security) so hopefully it's clearer now that you do one or the other.

lcawl:
I've decided to split this tutorial into multiple pages (i.e. more like the machine learning tutorial). 

@tvernum I also added a tip recommending that users explicitly set the xpack.security.enabled setting.

lcawl:
Backported to 6.x, 6.4, and 6.3 branches via https://github.com/elastic/stack-docs/commit/335e778bf37aa68b2d767800a007cdbf4f69884b

PR #89, opened 2018-07-25
Document that field-level security should not be set on field aliases.

jtibshirani:


jtibshirani:
Pinging @elastic/es-security

PR #98, opened 2018-08-07
[Kerberos] Add stack documentation for Kerberos realm

bizybot:
This commit adds documentation for Kerberos realm in
stack-docs. It briefly talks about Kerberos authentication,
its support and then points to settings and configuration
documentation.


docs/en/stack/security/authentication/kerberos-realm.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+
+To authenticate users with Kerberos, you need to
+{ref}/configuring-kerberos-realm.html[configure a Kerberos realm] and
+<<mapping-roles, map users to {security} roles>>.
+For more information on realm settings, see
+{ref}/security-settings.html#ref-kerberos-settings[Kerberos realm settings].
+
+[[kerberos-terms]]
+==== Key concepts
+
```
lcawl: I added this content here instead of https://github.com/elastic/elasticsearch/pull/32662, since it seems like a better fit with the overview info. If you disagree, let me know!

docs/en/stack/security/authentication/kerberos-realm.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+tickets. It can be used to identify a user or a service provided by a
+server.
++
+--
+Kerberos V5 principal names are of format `primary/instance@REALM`, where
+`primary` is a user name.
+
+`instance` is an optional string that qualifies the primary and is separated
+by a slash(`/`) from the primary. For a user, usually it is not used; for hosts, 
+it is the fully qualified domain name of the host.
```
lcawl: It is unclear why hosts are referenced here. Is that related to service principals?

docs/en/stack/security/authentication/kerberos-realm.asciidoc
bizybot: I wasn't sure about this when I added to the configuring Kerberos section as most of the action was there. But looking at it now makes sense here as it is an introduction part and helps to understand key concepts before moving on to the configuration part. Thank you.

docs/en/stack/security/authentication/kerberos-realm.asciidoc
bizybot: Not necessary but usually host is for service principals. Even users can have host FQDN as instance part of a principal name. The definition is as per MIT Kerberos documentation and I prefer to keep it the same. Thank you.

PR #119, opened 2018-09-10
[DOCS] Update monitoring diagrams with metricbeat

lcawl:
Related to https://github.com/elastic/beats/issues/7035

This PR replaces the architectural diagrams with one that includes Metricbeat. It also adds text to describe the beta behaviour.


lcawl:
You can preview the changes here: https://doc-review-63bf6.firebaseapp.com/x/how-monitoring-works.html

docs/en/stack/monitoring/how-monitoring-works.asciidoc
```diff
@@ -2,32 +2,37 @@
...
 but the versions cannot differ by more than one major version.
 
-The following diagram illustrates a typical monitoring architecture with
-separate production and monitoring clusters:
-
-image::monitoring/images/architecture1.jpg[A typical monitoring environment]
-
-If you have at least a Gold license, you can route data from multiple production
-clusters to a single monitoring cluster:
+If you are not using {metricbeat} to collect the {kib} data, you might want to
```
ycombinator: I wouldn't say there is a causality between the usage of metricbeat and having a dedicated kibana instance. I think the previous wording is more accurate over here:

> If Kibana is a regular part of your stack, you might want to create a dedicated Kibana instance for monitoring, rather than using a single Kibana instance to access both your production cluster and monitoring cluster:

docs/en/stack/monitoring/how-monitoring-works.asciidoc
```diff
@@ -2,32 +2,37 @@
...
+In general, the {es} cluster you are monitoring controls where the monitoring 
+metrics for the stack are stored. By default, they are stored in local indices. 
+
+In production, we strongly recommend using a separate monitoring cluster. Using 
+a separate monitoring cluster prevents production cluster outages from impacting
 your ability to access your monitoring data. It also prevents monitoring
-activities from impacting the performance of your production cluster.
+activities from impacting the performance of your production cluster. 
+
+beta[] In 6.5 and later, you can use {metricbeat} to collect and ship data about
```
ycombinator: I like that we are flagging this as a "beta" section, including the diagram that follows it. What do you think about also keeping the original non-beta way of doing things, including the diagram? Do you think it might be too much information and could lead to confusion?

docs/en/stack/monitoring/how-monitoring-works.asciidoc
lcawl: Though when you're using Kibana with Metricbeat, you're not accessing the production cluster, are you?

ycombinator: In a setup where Metricbeat is being used to monitor Kibana:
- Kibana has two connections: one to the production ES cluster and another to the Monitoring ES cluster. Of course, the two clusters might actually be the same, single cluster.
- Metricbeat has one connection: to the monitoring ES cluster.

Does this answer your question or did I just make it more confusing :)?

docs/en/stack/monitoring/how-monitoring-works.asciidoc
```diff
@@ -2,32 +2,39 @@
...
-
-image::monitoring/images/architecture2.jpg[A monitoring environment with multiple production clusters]
-
-If {kib} is a regular part of your stack, you might want to create a dedicated
-{kib} instance for monitoring, rather than using a single {kib} instance to
+If {kib} is a regular part of your stack, you might want to create a dedicated 
+{kib} instance for monitoring, rather than using a single {kib} instance to 
 access both your production cluster and monitoring cluster:
 
 image::monitoring/images/architecture3.jpg[A monitoring environment with separate {kib} instances]
```
ycombinator: Looking at the preview URL, this image only shows one Kibana instance. Maybe this ought to reference a different image file?

docs/en/stack/monitoring/how-monitoring-works.asciidoc
lcawl: Thanks, I've updated that paragraph, since it hadn't been clear to me what was meant by being "a regular part of the stack".

docs/en/stack/monitoring/how-monitoring-works.asciidoc
lcawl: The last image has the write and read coming from separate "instances" within the Kibana box, but if that's not clear, I can add a second Kibana box.

PR #135, opened 2018-10-26
Troubleshooting SAML doc

jkakavas:
Add a troubleshooting document for the SAML realm with common errors we've seen in the last months.

Resolves https://github.com/elastic/elasticsearch/issues/34690


jkakavas:
Kindly pinging for some :eyes: @lcawl @tvernum

tvernum:
Sorry @jkakavas I tend to rely on the GitHub UI to find outstanding PRs to review, and I forgot to check this repo last week.

docs/en/stack/security/troubleshooting.asciidoc
```diff
@@ -392,6 +392,123 @@ Kerberos/SPNEGO debug logging on JVM, add following JVM system properties:
...
+
+*Resolution:*
+
+. This can be resolved by correcting your `sp.acs` setting in your {es} SAML
+realm configuration or by using the `xpack.security.public.*` settings in `kibana.yml`.
+{kib} either uses the `server.host` and `server.port` or the `xpack.security.public.*`
+settings to construct the SAML ACS URL and this URL needs to match exactly the URL that
+is configured in the SAML realm configuration.
+
+NOTE: The two URLs are compared as case-sensitive strings and not as canonicalized URLs.
```
tvernum: I suspect there's not quite enough hand-holding here.
If someone is reading this, then the error message itself wasn't enough info for them, and I don't think this explanation will be either.
I think, at the very least we need a walk through of something like: _(bear in mind that it's just after midnight here, so you'll probably want to rewrite 90% of this)_
- The URL in the error message is the URL that Kibana thinks it should use for the "acs" 
- The URL in the `elasticsearch.yml` is what Elasticsearch thinks Kibana will use
- If you're getting this error then those aren't the same, but the need to be (as case-sensitive strings ...)
- Often it is as simple as changing the config in elasticsearch to match the ASC URL that Kibana is using (the one from the error message).
- However, if you think that the URL that Kibana is sending is wrong (e.g. Kibana is behind a proxy/load balancer with a different URL, or is accessed via a custom DNS name), then you can reconfigure Kibana so that it is aware of the correct URL:
    1. you can change `server.host` and `server.port` to cause Kibana to listen on a different URL. This may affect the URLs that users need to use to access Kibana, so you should consult the Kibana documentation for more details.
    2. You can configure Kibana to be aware that its _public_ URL is not the same as the server and port it listens on. The `xpack.security.public.*` settings are used to construct the ACS URL without changing the server and port that Kibana listens on.
- In either case, bear in mind that your IdP needs to be configured with this URL as well, so whatever URL you end up with in your `elasticsearch.yml` will need to match what you put in your IdP.

docs/en/stack/security/troubleshooting.asciidoc
```diff
@@ -392,6 +392,123 @@ Kerberos/SPNEGO debug logging on JVM, add following JVM system properties:
...
+. Authentication in Kibana fails and the following error is printed in the
+  Elasticsearch logs:
++
+--
+[source, shell]
+-----------------------------------------------
+Cannot find any matching realm for [SamlPrepareAuthenticationRequest{realmName=null,
+assertionConsumerServiceURL=https://my.kibana.url/api/security/v1/saml}]
+-----------------------------------------------
+--
```
tvernum: I would like the **Symptoms** and **Resolutions** to be paired in the document rather than being 2 lists.

docs/en/stack/security/troubleshooting.asciidoc
```diff
@@ -392,6 +392,123 @@ Kerberos/SPNEGO debug logging on JVM, add following JVM system properties:
...
+{kib} either uses the `server.host` and `server.port` or the `xpack.security.public.*`
+settings to construct the SAML ACS URL and this URL needs to match exactly the URL that
+is configured in the SAML realm configuration.
+
+NOTE: The two URLs are compared as case-sensitive strings and not as canonicalized URLs.
+
+. We could not find the metadata for the SAML Entity ID `your:entity.id` in the metadata
+file (`metadata.xml`) that has been configured. Either the value of `idp.entity_id` setting
+in the {es} SAML realm configuration is wrong/misspelled, or the wrong metadata file has been
+configured in `idp.metadata.path`.
```
tvernum: I think this needs more - my hope is that I can just use this doc as an answer to forum posts, but as it stands it doesn't go quite far enough to get users over the hump.

I think we need a few more specific steps like:
- The `metadata.xml` file need to be from your IdP 
- The start of the metadata file should have an element like below.  The value of the `entityID` attribute should be configured as the `idp.entity_id` in `elasticsearch.yml`
    ```
    <EntityDescriptor ID="0597c9aa-e69b-46e7-a1c6-636c7b8a8070" entityID="https://saml.example.com/f174199a-a96e-4201-88f1-0d57a610c522/" ...
    ``` 
- This is a case-sensitive string comparison ...

docs/en/stack/security/troubleshooting.asciidoc
```diff
@@ -392,6 +392,123 @@ Kerberos/SPNEGO debug logging on JVM, add following JVM system properties:
...
+
+. This error indicates that {es} failed to process the incoming SAML authentication message.
+Since the message can't be processed, {es} is not aware of who the to-be authenticated
+user is and the `<unauthenticated-saml-user>` placeholder is used instead. To diagnose the
+_actual_ problem one needs to check the {es} logs for further details.
+
+. This means that the SAML Identity Provider failed to authenticate the user and sent a
+SAML Response to the Service Provider (Elastic Stack) indicating this failure. The `Code`
+can be a different error identifier than the one above and might offer an indication about
+the cause of the failure, but `Message` and `Detail` will be `null`. In order to further
```
tvernum: > `Message` and `Detail` will be `null`.

They might not be (but usually are).
If they're `null`, then that means we have no additional information, but it's possible for the IdP to populate them.

I think we want to say something more like:
- If the IdP provided details about the cause of the error, then this will be included in the `Message` or `Detail` attributes of the log message.
- If these attributes are `null`, then the IdP did not provide additional details and you will need to consult the logs and/or documentation of the Identity Provider.

docs/en/stack/security/troubleshooting.asciidoc
```diff
@@ -392,6 +392,123 @@ Kerberos/SPNEGO debug logging on JVM, add following JVM system properties:
...
+the Identity Provider sent. {es} uses the public key of the Identity Provider that is
+included in the SAML metadata, in order to validate the signature that the IdP has created
+using its corresponding private key. Failure to do so, can have a number of causes:
+.. As the error message indicates, the most common cause is that the wrong metadata file
+is used and as such the public key it contains, doesn't correspond to the private key the
+Identity Provider uses.
+.. The configuration of the Identity Provider has changed or the key has been rotated and
+the metadata file that {es} is using has not been updated.
+.. The SAML Response has been altered in transit and the signature cannot be validated even
+if the correct key is used.
```
tvernum: Is it worth being clear that these certificates have nothing to do with SSL?
I'm not sure exactly what to say, but users (understandably) see cert/key in the config and think SSL and then tie themselves in knots.

docs/en/stack/security/troubleshooting.asciidoc
```diff
@@ -392,6 +392,123 @@ Kerberos/SPNEGO debug logging on JVM, add following JVM system properties:
...
+`authProvider` in {kib}. The process is documented in the <<saml-kibana-basic, SAML Guide>>
+
+Very detailed trace logging can be enabled specifically for the SAML realm by adding the
+following lines to the end of the `log4j2.properties` configuration file in the `ES_PATH_CONF`:
+
+[source,properties]
+----------------
+logger.saml.name = org.elasticsearch.xpack.security.authc.saml
+logger.saml.level = TRACE
+----------------
```
tvernum: It's really just personal preference, but I recommend adding them to the `elasticsearch.yml` instead, and never touching `log4j.properties`
```
logger.org.elasticsearch.xpack.security.authc.saml: TRACE
```

jkakavas:
Thanks for the feedback Tim ! I might have erred a little on the not adding enough details side. I took another swing at it

jkakavas:
Ping @tvernum for another look. @lcawl can you also chime in for the "issue - resolution" couples format (as it's different from the rest of the troubleshooting pages )

jkakavas:
@lcawl can we get some :eyes: on this please ? :)

jkakavas:
My weekly kind reminder @lcawl :)

lcawl:
Sorry for the delay, @jkakavas !  I fixed a broken link and updated the formatting.  You can preview the changes here: https://doc-review-63bf6.firebaseapp.com/x3/trb-security-saml.html

jkakavas:
Thanks for the review and the fixes @lcawl !

PR #148, opened 2018-11-09
[DOCS] Adds machine learning limitations related to Kibana spaces

lcawl:
This PR documents limitations related to the use of Kibana spaces with machine learning objects.

### DIFF
--- a/docs/en/stack/ml/limitations.asciidoc
+++ b/docs/en/stack/ml/limitations.asciidoc
@@ -225,8 +225,11 @@ the {ml-features} identify the file format and field mappings. You can then
 optionally import that data into an {es} index and create an index pattern. This 
 index pattern belongs to the space that is active when you import the file. If 
 you want to use the index pattern in other spaces, you must create it as needed.  
-* Likewise, if you clone a {ml} job in a space that does not contain the 
-necessary index pattern, the cloned job fails. 
+* Likewise, if you use the single-metric, multi-metric, or population job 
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/limitations.asciidoc
```diff
@@ -208,3 +208,30 @@ pattern that uses the Rollup feature, the {ml} job creation wizards fail.
...
+index patterns, dashboards, and visualizations) that might not be available in 
+all spaces. For example:
+
+* experimental[] If you upload a file on the *Machine Learning* page in {kib}, 
+the {ml-features} identify the file format and field mappings. You can then 
+optionally import that data into an {es} index and create an index pattern. This 
+index pattern belongs to the space that is active when you import the file. If 
+you want to use the index pattern in other spaces, you must create it as needed.  
+* Likewise, if you clone a {ml} job in a space that does not contain the 
+necessary index pattern, the cloned job fails.
```
pheyos: To be precise, this only happens if the original job was created with the Single Metric, Multi Metric or Population Wizard. Cloning jobs created via the Advanced Wizard or the API works. And technically, it's not the cloned job that fails, but the (Single Metric, Multi Metric, Population) Wizard fails to load.
Not sure if we want to be that detailed.

pheyos: The changes are looking good @lcawl :+1:

PR #153, opened 2018-11-19
[DOCS] Kerberos troubleshooting documentation for problems with curl

bizybot:
Kerberos realm only supports Spnego mechanism (Oid 1.3.6.1.5.5.2) and
does not yet support Kerberos mechanism (Oid 1.2.840.113554.1.2.2).
Customers usually start initial testing using `curl` to test
Elasticsearch - Kerberos integration and may face problems
authenticating users using Kerberos.

Usually, the problems are due to:
- Known issues with older versions of `curl`.
It is recommended to use `curl` version 7.49 or above as earlier
versions are known to have issues
(like https://bugzilla.redhat.com/show_bug.cgi?id=1333004)
- Sometimes the support for `GSS-API`, `Kerberos` and `SPNEGO`
features are missing from the curl. One will need to download and install
the appropriate version of curl where these features exist or need to
compile with the required support.

This commit adds the troubleshooting documentation to help customers
when they face these issues and provide a way to resolve by installing
the appropriate version of curl with features that need to be supported.

### DIFF
--- a/docs/en/stack/security/troubleshooting.asciidoc
+++ b/docs/en/stack/security/troubleshooting.asciidoc
@@ -392,6 +392,30 @@ Kerberos/SPNEGO debug logging on JVM, add following JVM system properties:
 
 For more information about JVM system properties, see {ref}/jvm-options.html[configuring JVM options].
 
+`gss_init_sec_context() failed: An unsupported mechanism was requested`::
+
+`No credential found for: 1.2.840.113554.1.2.2 usage: Accept`::
++
+--
+
+You would usually see this error message on the client side when using `curl` to test {es} Kerberos
+setup, as the support for Kerberos Spnego is missing on the client due to an old version of curl.
+Kerberos realm in {es} only supports Spengo mechanism (Oid 1.3.6.1.5.5.2) and does not yet support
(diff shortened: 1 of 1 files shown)


lcawl:
Thanks for writing this, @bizybot !  I moved the new content up above the more generic info about tracing, hope that's okay.

docs/en/stack/security/troubleshooting.asciidoc
```diff
@@ -368,6 +368,31 @@ the time on the machines within the domain is in sync.
...
 
+`gss_init_sec_context() failed: An unsupported mechanism was requested`::
+
+`No credential found for: 1.2.840.113554.1.2.2 usage: Accept`::
++
+--
+
+You would usually see this error message on the client side when using `curl` to
+test {es} Kerberos setup, as the support for Kerberos Spnego is missing on the
+client due to an old version of curl. The Kerberos realm in {es} only supports
```
lcawl: Would it be correct to reword this sentence as follows?:
"...Kerberos setup. For example, these messages occur when you are using an old version of curl on the client and therefore Kerberos Spnego support is missing".

PR #157, opened 2018-11-28
Add missing `host` and `cloud` metadata processors

exekias:
Infrastructure UI makes use of the data collected by these processors, they are enabled by default starting with Beats 6.5 but I think its good to add them to the list of required info for awareness.

### DIFF
--- a/docs/en/infraops/installation.asciidoc
+++ b/docs/en/infraops/installation.asciidoc
@@ -67,6 +67,8 @@ To populate the *Hosts* view and add logs, enable:
 * {filebeat-ref}/filebeat-module-system.html[{filebeat} `system` module]
 * {filebeat-ref}/filebeat-modules.html[Other {filebeat} modules] needed for
 your environment, such as `apache2`, `redis`, and so on
+* {metricbeat-ref}/add-host-metadata.html[{metricbeat} `add_host_metadata` processor]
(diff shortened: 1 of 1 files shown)


docs/en/infraops/installation.asciidoc
```diff
@@ -67,6 +67,8 @@ To populate the *Hosts* view and add logs, enable:
 * {filebeat-ref}/filebeat-module-system.html[{filebeat} `system` module]
 * {filebeat-ref}/filebeat-modules.html[Other {filebeat} modules] needed for
 your environment, such as `apache2`, `redis`, and so on
+* {metricbeat-ref}/add-host-metadata.html[{metricbeat} `add_host_metadata` processor]
```
dedemorton: Maybe you should indicate parenthetically that these are enabled by default? My concern is that users be frustrated if they spend time learning how to enable the processors but then find out that they are already enabled. 

so maybe do something like:

```
* {metricbeat-ref}/add-host-metadata.html[{metricbeat} `add_host_metadata` processor] (enabled by default)
* {metricbeat-ref}/add-cloud-metadata.html[{metricbeat} `add_cloud_metadata` processor] (enabled by default)
```

I think you could do the same for the metricbeat system module, if I'm not mistaken.

exekias: Good point @dedemorton, I'll make the changes

dedemorton:
@exekias If you want to assign this to me, I can take over the change when I get back after the holidays.

PR #160, opened 2018-12-03
Add section for misconfigured sp entity id

jkakavas:
This started popping up quite a lot in cloud, with the culprit
usually being trailing slashes in the SP entity ID that is set
in either `elasticsearch.yml` or the IdP configuration.


tvernum:
LGTM. (Can't officially review on mobile)

PR #181, opened 2019-01-12
Add links to Kibana docs

### LINKED ISSUE #180: Improve linking between Infrastructure Monitoring docs and Kibana docs
The topic about getting started with the Infrastructure UI needs to have a pointer to the Kibana documentation about UI settings:

https://www.elastic.co/guide/en/kibana/master/infrastructure-ui-settings-kb.html

In general, make sure there are sufficient pointers between the topics.

dedemorton:
Closes #180

### DIFF
--- a/docs/en/infraops/infra-ui-intro.asciidoc
+++ b/docs/en/infraops/infra-ui-intro.asciidoc
@@ -8,6 +8,9 @@ After you have infrastructure monitoring <<install-infrastructure-monitoring,up 
 and data is streaming to {es}, use the {infra-ui} UI in {kib} to monitor your
 infrastructure and identify problems in real time.
 
+For more information about working with the {infra-ui} UI, see the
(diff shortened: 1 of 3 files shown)


tbragin:
LGTM!

dedemorton:
> It would be nice to have ability to duplicate key snippets, such as specifically how to customize Logs UI and Infrastructure UI index configuration right inline.

@tbragin Agreed. We do have a place in the docs repo where we can store content shared across books, but I've noticed that folks forget to update those files. The files don't show up in a repo search, so maintainers don't think to update them. So I hesitate to use that approach right now. I'd really like us to resolve the larger problem with our information architecture because the fragmentation of content is pervasive and not a great user experience. Having our content in silos organized by product is a legacy problem that's going to take a concentrated effort and investment to resolve. :-/ I'm looking forward to talking more with you about the issue, especially wrt the observability and beats docs.

PR #193, opened 2019-01-30
Fix broken link to ECE docs

nrichers:
This link from the glossary was broken after merging our ECE docs reshuffle.


nrichers:
I'm going to merge this, since the link broke the doc builds.

PR #197, opened 2019-02-02
[DOCS] Adds breaking changes page with content re-use

lcawl:
This PR adds a "Breaking Changes" page to the Installation and Upgrade Guide (https://www.elastic.co/guide/en/elastic-stack/master/index.html) and lays the groundwork for adding highlighted breaking changes from across the stack to this book.

After we move to an Asciidoctor build for this book, we can use tagged regions to pull in content from  breaking change pages across the library (per https://asciidoctor.org/docs/user-manual/#by-tagged-regions). At that time, the build command will need to change as follows (with --resources parameters added for each repo we pull from):

 docbldstk  --resource=elasticsearch/docs --resource=kibana/docs --asciidoctor

NOTE: The presence of these tagged regions cannot be processed by our current asciidoc builds, thus the following line must be added to the index.asciidoc to preview the tagged regions:

> :asciidoctor-builds: true

### DIFF
--- a/docs/en/install-upgrade/breaking.asciidoc
+++ b/docs/en/install-upgrade/breaking.asciidoc
@@ -1,0 +1,31 @@
+[[elastic-stack-breaking-changes]]
+== Breaking changes
+
+Before you upgrade, you must review the breaking changes for each product you
+use and make the necessary changes so your code is compatible with {version}. 
+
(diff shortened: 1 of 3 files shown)


lcawl:
You can preview this information in https://doc-review-63bf6.firebaseapp.com/x4/elastic-stack-breaking-changes.html

PR #194, opened 2019-01-31
[DOCS] Add OS/architecture classifier to distributions

lcawl:
Related to https://github.com/elastic/elasticsearch/pull/37881

This PR updates the installation package names in the stack-docs repo. For example, it adds a new Linux section to https://www.elastic.co/guide/en/elastic-stack-get-started/master/get-started-elastic-stack.html#install-elasticsearch

PR #198, opened 2019-02-05
Add linux examples to the stack getting started

dedemorton:
Adds Linux example commands to getting started docs. This is meant to complement https://github.com/elastic/stack-docs/pull/194.

Question: I'm not sure why we use the linux distros for Deb and RPM for Kibana. Maybe for historical reasons? Should we change this to show steps for using the Deb/RPM packages and starting Kibana as a service?


lcawl:
@dedemorton Maybe that's a mistake?  There are appropriate instructions in https://www.elastic.co/guide/en/kibana/6.6/deb.html for example

dedemorton:
@lcawl Might have been intentional to keep things simple? The steps do work as documented, so they are correct, if not ideal. I don't have time to test new commands atm, so maybe we can get this merged and follow up later with another PR?

PR #219, opened 2019-02-15
[DOCS] Remove upgrade step that stops machine learning

lcawl:
This PR removes the step related to stopping machine learning in the Installation and Upgrade Guide (https://www.elastic.co/guide/en/elastic-stack/master/upgrading-elastic-stack.html). 

Related to https://github.com/elastic/elasticsearch/pull/38876


droberts195:
It's not related to this PR but I noticed the file that's changed contains "IMPORTANT: 2.x indices are not compatible with {version}".  The "2.x" in that bit should be whatever is two major versions earlier, so for 8.x it should be 6.x and for 7.x it should be 5.x.

docs/upgrading-stack.asciidoc
```diff
@@ -68,10 +68,6 @@ restart all nodes to maintain communication across the cluster.
 password (`changeme`). For more information, see
 {stack-ov}/built-in-users.html[Built-in users].
 
-. Stop any {ml} jobs that are running before starting the
-upgrade process. See
-{stack-ov}/stopping-ml.html[Stopping {ml}].
```
droberts195: Maybe instead of completely removing this section it could say something like "Consider stopping any {ml} jobs that are running - see [<close_ml.asciidoc>](https://github.com/elastic/elasticsearch/pull/38876/files#diff-131043b47d93bd06a9c3a94c8a6d7624) for the pros and cons of doing this".

Or do you think that's just too complicated for this guide?

lcawl: Thanks, I've altered that text.

lcawl:
> It's not related to this PR but I noticed line 26 of the file that's changed contains...

There are quite a few of those sprinkled around, which we'll address in separate PRs.

PR #220, opened 2019-02-19
[DOCS] Update version detail in upgrade instructions

lcawl:
This PR updates the "important" warning at the top of https://www.elastic.co/guide/en/elastic-stack/master/upgrading-elastic-stack.html

Per https://github.com/elastic/stack-docs/pull/219#issuecomment-465162904

> ... The "2.x" in that bit should be whatever is two major versions earlier, so for 8.x it should be 6.x and for 7.x it should be 5.x.

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -23,8 +23,8 @@ Rolling upgrades are supported when upgrading from Elasticsearch 5.6 and
 Elasticsearch 6.0-6.2 to {version}. Upgrading from any
 version prior to 5.6 requires a full cluster restart.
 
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -23,8 +23,8 @@ Rolling upgrades are supported when upgrading from Elasticsearch 5.6 and
 Elasticsearch 6.0-6.2 to {version}. Upgrading from any
 version prior to 5.6 requires a full cluster restart.
 
-IMPORTANT: 2.x indices are not compatible with {version}. You must
-remove or reindex them on your 5.n cluster before upgrading to {version}. The internal
+IMPORTANT: 6.x indices are not compatible with {version}. You must
```
bleskes: 6.x or older?

PR #227, opened 2019-02-25
[DOCS] Fixes anchor for asciidoctor

lcawl:
This PR fixes a problem with the numbering in a list in the Getting Started Guide when it's built via asciidoctor:

image

Per https://asciidoctor.org/docs/user-manual/#anchordef this problem can be fixed by using an inline anchor


nik9000:
Lgtm! Thanks!

PR #231, opened 2019-03-04
[DOCS] Drafts data frame privileges

lcawl:
Related to https://github.com/elastic/elasticsearch/pull/39661

This PR drafts definitions for the `manage_data_frame_transforms` and `monitor_data_frame_transforms` cluster privileges and the `data_frame_transforms_admin` and 
`data_frame_transforms_user` built-in roles.

### DIFF
--- a/docs/en/stack/security/authorization/built-in-roles.asciidoc
+++ b/docs/en/stack/security/authorization/built-in-roles.asciidoc
@@ -37,6 +37,14 @@ suitable for writing beats output to {es}.
 
 --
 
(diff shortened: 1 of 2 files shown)


hendrikmuhs:
Data Frame (transforms) isn't part of ML, only Data Frame Analytics will be. Data Frame (transforms) is it's own plugin. 

FWIW: Data Frame Analytics will have it's own roles and we think about adding roles to combine the transforms and analytics one.

Issue #226, opened 2019-02-25
7.0 stack upgrade docs feedback


cjcenizal:
## Overview

As @bleskes coordinates testers for the 7.0 stack upgrade process, we'll gather feedback on the docs and record it here.

* The page we're using to guide us: https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elastic-stack.html
* Notes for testers: https://docs.google.com/spreadsheets/d/1LtmjfCgtohS_H6fMk3fSKh3tmEqyjioZbFk2giFDeQA/edit#gid=590440521

cjcenizal:
## "Upgrade Elasticsearch" docs

I think we should begin these docs with a single section titled "Upgrading the Elastic Stack?" which explains that if you're upgrading the Elastic Stack to visit the "Upgrade Elastic Stack" docs first. By giving this section its own title we can separate it from the rest of the other content, and ES-only upgraders can ignore it more easily. Might it also be worth removing the link to the Interactive Upgrade Guide, to avoid giving the user too many options?

## Backups

The first step of preparing to upgrade is to back-up your data, which links to [Snapshot and Restore docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/modules-snapshots.html#_shared_file_system_repository). These docs attempt to explain how to define where ES will create snapshots on the filesystem:

> Assuming that the shared filesystem is mounted to /mount/backups/my_fs_backup_location...

I find this wording a bit confusing because it assumes the reader knows a) what makes a filesystem "shared" and b) what it means to have a shared filesystem "mounted" somewhere. I used this [third party tutorial](https://qbox.io/blog/elasticsearch-data-snapshots-restore-tutorial) to learn what this meant. I'd suggest rephrasing this line to be:

> Assuming you want to store your snapshots inside of the directory at the absolute path /backups/my_fs_backup_location...

And if we have an official tutorial on snapshot and restore, I recommend we link directly to it from the docs.

## Deprecations

The link to the deprecation log doesn't seem to take you to a useful location... it's pointed at https://www.elastic.co/guide/en/elasticsearch/reference/6.6/settings.html, but I'm not sure how that relates?

## Breaking changes

The link for Elasticsearch should point to the version-specific page.

bleskes:
Thx for opening this @cjcenizal.

droberts195:
> what makes a filesystem "shared"

[...]

PR #234, opened 2019-03-07
[DOCS] Adds placeholders for version-specific upgrade steps

lcawl:
Related to https://github.com/elastic/stack-docs/issues/226

This PR moves the existing "Upgrading from 5.6" content to a separate file and creates placeholders for equivalent pages for 6.x- and 7.x-specific instructions.

It also switches the Installation and Upgrade Guide to pulling its version information from the shared versions files, so that it can use the new "prev-major-version" attribute in https://github.com/elastic/docs/pull/675.


lcawl:
Per discussion today, these version-specific sections might just contain links to the appropriate version of the documentation.  That is to say, only the upgrade path from the most recent version will be covered in detail.

lcawl:
This PR is replaced by https://github.com/elastic/stack-docs/pull/242

PR #235, opened 2019-03-07
[DOCS] Clarifies how to upgrade internal indices

lcawl:
Related to https://github.com/elastic/stack-docs/issues/226
Dependent on https://github.com/elastic/docs/pull/675

This PR clarifies the following section of the Installation and Upgrade Guide:
https://www.elastic.co/guide/en/elastic-stack/master/upgrading-elastic-stack.html#upgrade-internal-indices

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -227,8 +227,9 @@ compatible index structure.
 
 If you are upgrading from a version prior to {prev-major-version}, you must
 upgrade to the most recent version of {prev-major-version} first. Then you can
-use the {kib} Upgrade Assistant or the
-{ref}/docs-reindex.html[reindex API] to reindex the internal indices. 
+use the {kibana-ref}/upgrade-assistant.html[{kib} Upgrade Assistant] or the
+{ref}/docs-reindex.html[reindex API] to reindex the internal indices. For more
+information, see {ref}/reindex-upgrade.html[Reindex before upgrading].
 
 [TIP]
 ====
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -215,7 +215,7 @@ modifications to the document data and metadata during reindexing.
 index as alias for the new index and delete the 2.x index.
 
 [[upgrade-internal-indices]]
-==== Upgrading internal indices for {major-version}
+=== Upgrading internal indices for {major-version}
 
 {es} can read indices that were created in the *previous major version*. Older
 indices must be reindexed or deleted. For example, {es} {major-version} can use
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -223,37 +223,49 @@ indices created in {es} {prev-major-version}, but not those created in earlier
 versions. This restriction also applies to the internal indices that are used by
 Kibana and the {xpack} features. Therefore, before you can use Kibana and
 {xpack} features in {version}, you must upgrade the internal indices to a
-compatible index structure. If you are upgrading from a version prior to
-{prev-major-version}, you must upgrade to the most recent version of
-{prev-major-version} first. Then you can use the {kib} Upgrade Assistant or
-equivalent {ref}/migration-api-assistance.html[`_migration/assistance` API] to
-check whether you need to upgrade the internal indices. 
+compatible index structure.
 
-////
-//TBD
-To upgrade the `.security` index:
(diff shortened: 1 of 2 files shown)


lcawl:
In case it helps, you can preview the pertinent page here: https://doc-review-63bf6.firebaseapp.com/x/upgrading-elastic-stack.html#upgrade-internal-indices

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -215,45 +215,57 @@ modifications to the document data and metadata during reindexing.
...
-the {ref}/migration-api-assistance.html[`_migration/assistance` API]:
-
-[source,json]
+=== Upgrading internal indices for {major-version}
+
+{es} can read indices that were created in the *previous major version*. Older
+indices must be reindexed or deleted. For example, {es} {major-version} can use
+indices created in {es} {prev-major-version}, but not those created in earlier
+versions. This restriction also applies to the internal indices that are used by
+Kibana and the {xpack} features. Therefore, before you can use Kibana and
```
bleskes: I believe that Kibana reindexes it's own index these days. @epixa can you confirm?

epixa: We do, but if someone was attempting to upgrade from something like 6.2 to 7.x, then they wouldn't have run saved object migrations yet, so their index could have originated back in 5.6. So they'll still need to run something like 6.7 first to ensure the index is migrated.

We also don't delete the old indices automatically, we just create new ones for migrations. Does ES fail to start if it has old indices or does it just not recognize those indices?

Finally, this blurb of text would still mostly make sense even with Kibana doing migrations of its own index. We indirectly rely on ES things like the security index, so if that needs to be functional in order for Kibana (with security) to be accessible.

bleskes: Thanks @epixa . 

> We also don't delete the old indices automatically, we just create new ones for migrations. Does ES fail to start if it has old indices or does it just not recognize those indices?

Yes. ES will refuse to start. 

> but if someone was attempting to upgrade from something like 6.2 to 7.x, then they wouldn't have run saved object migrations yet, so their index could have originated back in 5.6

This implies that they'll do a full cluster restart, which will also mean that their ES 7.0 will indeed fail to start and they will be forced to downgrade and fix it.

@lcawl part of the confusion is caused by the usage of the word "upgrade" - in 5.x we had a dedicated upgrade API that people needed to activate. With 6.x it's a vanilla reindex via the reindex API (or the migration assistant UI). I think we should change the doc a bit to reflect this.

lcawl: @bleskes Do you mean that in these steps we should be advising users to call the reindex API (https://www.elastic.co/guide/en/elasticsearch/reference/7.0/docs-reindex.html) instead of the migration upgrade API
(https://www.elastic.co/guide/en/elasticsearch/reference/7.0/migration-api-upgrade.html)?

bleskes: > Do you mean that in these steps we should be advising users to call the reindex API instead of the migration upgrade API

We should tell people that the index needs to be reindexed. That can be done via the migration assistant UI or alternatively directly via the reindex API (In that order ;))

lcawl: Thanks @bleskes !  I've updated the text to reflect that. Should we also update the documentation for the migration upgrade API (https://www.elastic.co/guide/en/elasticsearch/reference/master/migration-api-upgrade.html) to recommend that they use the reindex API instead?

lcawl:
This PR will be replaced by https://github.com/elastic/elasticsearch/pull/40019 since this section is disappearing from the Installation and Upgrade Guide in https://github.com/elastic/stack-docs/pull/242

docs/en/install-upgrade/upgrading-stack.asciidoc
bleskes: Thx @lcawl . Re the [upgrade api](https://www.elastic.co/guide/en/elasticsearch/reference/master/migration-api-upgrade.html) - I had a quick chat with @gwbrown about what to do with them, I *think* he was taking it up to the team to decide. @gwbrown , do I recall correctly?

gwbrown: Yes, we should also update the docs for the Migration Upgrade API to point users towards the Kibana Upgrade Assistant - we're also going to be deprecating the Migration Assistance and Migration Upgrade APIs in 6.7 and removing in 7.0, see https://github.com/elastic/elasticsearch/issues/40014.  I'll have PRs up for that later today.

lcawl: Thanks!  I also implemented those changes in https://github.com/elastic/stack-docs/pull/247

PR #242, opened 2019-03-12
7.0 upgrade

debadair:
First pass at the upgrade changes for 7.0.

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -75,9 +75,28 @@ and other enhancements.
 [[oss-stack-upgrade]]
 === Upgrading from 6.6 or earlier
 
-To upgrade directly to {version} from 6.6 or earlier, you must
+To upgrade directly to {es} {version} from versions 6.0-6.6, you must
 {ref}/reindex-upgrade.html[manually reindex] any 5.x indices you need to
-carry forward and perform a {ref}/restart-upgrade.html[full cluster restart].
+carry forward, and perform a {ref}/restart-upgrade.html[full cluster restart].
+This includes any internal indices created in 5.x, such as the `.kibana` and
+`.security*` indices.
+
+[TIP]
+====
+Add a temporary superuser account in the `file` realm before reindexing the
+internal `.security*` indices by running the
+{ref}/users-command.html[elasticsearch-users useradd] command on a single node.
+This provides a backup authentication method if there is a problem during
+reindexing.
+====
+
+Make sure all 5.x indices have been deleted before upgrading to {version}. {es}
+{version} will fail to start if any 5.x indices are present.
+
+If you are running a version prior to 6.0,
+https://www.elastic.co/guide/en/elastic-stack/6.7/upgrading-elastic-stack.html[upgrade to 6.7]
+and reindex your old indices or bring up a new {version} cluster and
+<<reindex-upgrade-remote, reindex from remote>>.
 
 The recommended path is to upgrade to 6.7 before upgrading to {version}. This
 makes it easier to identify the changes you need to make to upgrade and enables
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -24,7 +24,7 @@ By default, deprecation warnings are logged when the log level is set to `WARN`.
 . Review the <<elastic-stack-breaking-changes>> and upgrade your code to work
 with {version}.
 
-. Upgrade to 6.7 and use the {kib} Upgrade Assistant to {ref}/docs-reindex.html[reindex]
+. Upgrade to 6.7 and use the {kibana-ref}/upgrade-assistant.html[{kib} Upgrade Assistant] to {ref}/docs-reindex.html[reindex]
 any indices that are not compatible with {version}.
 +
 [role="xpack"]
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -1,277 +1,105 @@
...
-<<upgrade-internal-indices,use the `_migration/upgrade` API>> directly to
-upgrade the internal indices after you install Elasticsearch {version}.
+. Review the <<elastic-stack-breaking-changes>> and upgrade your code to work
+with {version}.
 
-. If you use {stack} {security-features} to secure your cluster:
-.. Make sure TLS is enabled to encrypt communications between nodes. TLS must
-be enabled to upgrade to {version}. For more information, see
-{stack-ov}/encrypting-communications.html[Encrypting communications].
+. Upgrade to 6.7 and use the {kib} Upgrade Assistant to {ref}/docs-reindex.html[reindex]
```
lcawl: Since reindex is a link to the API maybe we should add a link to the UI info too?

```suggestion
. Upgrade to 6.7 and use the {kibana-ref}/upgrade-assistant.html[{kib} Upgrade Assistant] to {ref}/docs-reindex.html[reindex]
```

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -1,277 +1,105 @@
...
---
-
-. Delete the temporary superuser account from the file realm.
+=== Upgrading from 6.6 or earlier
 
-You can use your regular administration credentials to upgrade the other
-internal indices using the `_migration/upgrade` API.
+To upgrade directly to {version} from 6.6 or earlier, you must
+{ref}/reindex-upgrade.html[manually reindex] any 5.x indices you need to
+carry forward and perform a {ref}/restart-upgrade.html[full cluster restart].
```
lcawl: It might also be an option to create a new 7.x cluster and reindex from that cluster per https://www.elastic.co/guide/en/elasticsearch/reference/7.0/reindex-upgrade-remote.html ?

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -1,277 +1,105 @@
...
```
lcawl: Per https://github.com/elastic/stack-docs/pull/235
```suggestion
carry forward (including the internal indices that are used by {kib} and the {xpack} features) and perform a {ref}/restart-upgrade.html[full cluster restart]. 
```

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -1,277 +1,124 @@
...
-{xpack} Reindex Helper UI to upgrade the other indices.
+=== Upgrading from 6.6 or earlier
+
+To upgrade directly to {es} {version} from versions 6.0-6.6, you must
+{ref}/reindex-upgrade.html[manually reindex] any 5.x indices you need to
+carry forward, and perform a {ref}/restart-upgrade.html[full cluster restart].
+This includes any internal indices created in 5.x, such as the `.kibana` and
+`.security*` indices.
+
+[TIP]
```
lcawl: This TIP is being added to the Elasticsearch Reference via https://github.com/elastic/elasticsearch/pull/40019, so it can be omitted here.

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -1,272 +1,124 @@
...
-+
-If your trial license expires,
-https://register.elastic.co/[register for a free Basic license]. To apply the
-license, upload the license file with the `license` API:
-+
-[source,json]
-----------------------------------------------------------
-license -d @license.json
-----------------------------------------------------------
+. Back up your data. You **cannot roll back** to an earlier version unless
```
cjcenizal: I ran into a problem when I included the .kibana index in my snapshot. When I subsequently tried to restore this index it conflicted with the existing one. Can someone from @elastic/kibana-platform tell me if this was user error on my part or if it is expected?

If expected, then we should update the docs with guidance on avoiding this pitfall.

debadair: This feels like general snapshot/restore behavior--you can't restore on top of an open index. As far as the upgrade process goes, we tell folks to back up their data, but don't really get into how to use the snapshot if the upgrade goes sideways. (I made the implicit assumption that they would understand how to use the backup if they needed to.) It might make sense to add a few words when we add the info about what to do if you have a "broken upgrade".

cjcenizal: > This feels like general snapshot/restore behavior--you can't restore on top of an open index. 

Thanks Deb! That makes sense.

> It might make sense to add a few words when we add the info about what to do if you have a "broken upgrade".

Yes, that would be very helpful! Including a note about what this means for Kibana would also be helpful, I think. Something along the lines of, "Note for Kibana users: If you're restoring a snapshot that contains the .kibana index, you'll need to shut Kibana down, close the index, and restore using the command line".

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -1,272 +1,115 @@
...
-. If you disabled the {es} {security-features}, also disable the {kib}
-{security-features} in `kibana.yml`:
-+
-[source,yaml]
-----------------------------------------------------------
-xpack.security.enabled: false
-----------------------------------------------------------
+.. {es} Hadoop: {hadoop-ref}/install.html[install instructions]
+.. {es}: {ref}/setup-upgrade.html[upgrade instructions]
+.. Kibana: {kibana-ref}/upgrade.html[upgrade instructions]
```
cjcenizal: Per feedback from @tylersmalley, this is a little more complicated if you're using Kibana. Because Kibana comes with a new Upgrade Assistant which tells you if there are problems with your indices and cluster, the order should be:

1. Once the user is on 6.7, go into Kibana and view the Upgrade Assistant. Address all critical items.
2. Follow the ES upgrade steps.
3. Follow the Kibana upgrade steps.

For context, here's what the Upgrade Assistant looks like (this screenshot may contain bugs so it's not appropriate for use in the docs):

image

debadair:
Will address CJ's comments in a separate PR.

PR #247, opened 2019-03-15
[DOCS] Deprecates migration assistance and upgrade APIs

lcawl:
Related to https://github.com/elastic/elasticsearch/pull/40072

This PR adds deprecation notes to the migration assistance and migration upgrade APIs and suggests alternative methods.

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -202,9 +202,8 @@ You can also call the Elasticsearch migration APIs directly:
 `/_xpack/migration/deprecations`:: Retrieves information about cluster-, node-,
 and index-level settings that use deprecated features.
 +
-`/_xpack/migration/assistance`:: deprecated[6.7.0] Runs a series of checks on
-your cluster, nodes, and indices and returns a list of issues that need to be
-fixed before you can upgrade to {version}.
+`/_xpack/migration/assistance`:: deprecated[6.7.0] Returns a list of indices
+that need to be reindexed before you can upgrade to {version}.
 +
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -199,12 +199,16 @@ need to be migrated or deleted, and upgrade the internal indices to the
...
 You can also call the Elasticsearch migration APIs directly:
 +
-`/_xpack/migration/assistance`:: Runs a series of checks on your cluster,
-nodes, and indices and returns a list of issues that need to be
+`/_xpack/migration/deprecations`:: Retrieves information about cluster-, node-,
+and index-level settings that use deprecated features.
++
+`/_xpack/migration/assistance`:: deprecated[6.7.0] Runs a series of checks on
+your cluster, nodes, and indices and returns a list of issues that need to be
 fixed before you can upgrade to {version}.
```
gwbrown: This should probably say something like

```
`/_xpack/migration/assistance`:: deprecated[6.7.0] Returns a list of indices that need to 
be reindexed before you can upgrade to {version}.
```

As it doesn't actually check anything else - the description as written applies more to the `migration/deprecations` API.

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -240,33 +244,25 @@ version prior to 5.6, you must upgrade them after after installing
 Elasticsearch {version}.
 
 To get a list of the indices that need to be upgraded, install {xpack} and use
```
gwbrown: Do we still want to tell people to "install xpack"?

lcawl: Good point, I've removed that.

PR #248, opened 2019-03-20
[DOCS] Add 'recovery' to glossary for issue #7264

jrodewig:
Adds a definition for `recovery` to the Glossary. Closes /elastic/elasticsearch#7264.

### DIFF
--- a/docs/en/glossary/glossary.asciidoc
+++ b/docs/en/glossary/glossary.asciidoc
@@ -552,7 +552,7 @@ endif::cloud-terms[]
 ifdef::elasticsearch-terms[]
 
 [[glossary-recovery]] recovery ::
-The process of loading an index shard from disk, making the shard available for queries.
+The process of updating a shard copy and making it available for queries.
 +
 Recovery automatically occurs anytime a shard moves to a different node in the same cluster, including:
 
--- a/docs/en/glossary/glossary.asciidoc
+++ b/docs/en/glossary/glossary.asciidoc
@@ -552,7 +552,7 @@ endif::cloud-terms[]
 ifdef::elasticsearch-terms[]
 
 [[glossary-recovery]] recovery ::
-The process of updating a shard copy and making it available for queries.
+The process of syncing a shard copy from a source shard. Upon completion, the recovery process makes the shard copy available for queries.
 +
 Recovery automatically occurs anytime a shard moves to a different node in the same cluster, including:
 
--- a/docs/en/glossary/glossary.asciidoc
+++ b/docs/en/glossary/glossary.asciidoc
@@ -551,6 +551,20 @@ nodes handling the user requests.
 endif::cloud-terms[]
 ifdef::elasticsearch-terms[]
 
+[[glossary-recovery]] recovery ::
+The process of loading an index shard from disk, making the shard available for queries.
++
+Recovery automatically occurs anytime a shard moves to a different node in the same cluster, including:
+
+* Node startup
+* Node failure
+* Index shard replication
+* Snapshot restoration
++
+//Source: Elasticsearch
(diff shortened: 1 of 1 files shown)


docs/en/glossary/glossary.asciidoc
```diff
@@ -551,6 +551,20 @@ nodes handling the user requests.
 endif::cloud-terms[]
 ifdef::elasticsearch-terms[]
 
+[[glossary-recovery]] recovery ::
+The process of loading an index shard from disk, making the shard available for queries.
```
nik9000: We've been saying "shard copy" more than "index shard", mostly because it implies that each shard copy is pretty much the same.

I don't think "from disk" is super accurate. I might say something like "The process of bringing a shard copy up to date and making it available for queries." When we do it on startup we do it by recovering what we have locally and then pulling changes from other shard copies on other nodes. When we do it for replication we start the recovery with a "copy all the files" step. I guess *where* we are doing the recovery from is really interesting which is why I'm stuck on the "from disk" part.

jrodewig: Thanks @nik9000. I think I was too narrowly focused on startup there.

I'll submit another commit to update that sentence to `The process of updating a shard copy and making it available for queries.`

It's also makes it clearer that making the shard available is a step of recovery, not just an outcome.

docs/en/glossary/glossary.asciidoc
```diff
@@ -551,6 +551,20 @@ nodes handling the user requests.
 endif::cloud-terms[]
 ifdef::elasticsearch-terms[]
 
+[[glossary-recovery]] recovery ::
+The process of updating a shard copy and making it available for queries.
++
+Recovery automatically occurs anytime a shard moves to a different node in the same cluster, including:
+
```
nik9000: Does there have to be a `+` here? I can never tell without building it.

docs/en/glossary/glossary.asciidoc
```diff
@@ -551,6 +551,20 @@ nodes handling the user requests.
 endif::cloud-terms[]
 ifdef::elasticsearch-terms[]
 
+[[glossary-recovery]] recovery ::
+The process of updating a shard copy and making it available for queries.
```
nik9000: I think "updating" isn't super clear here. Neither was my "bring up to date". The idea is that recovery is about getting the on disk and in memory data structures "ready" and "current" based on data *somewhere*. On disk if this is the first node coming online with that shard copy. From the snapshot if you are restoring from a snapshot. Or from whatever node has the primary copy of the shard otherwise. I think that is all the cases, but this was never my area of expertise.

docs/en/glossary/glossary.asciidoc
jrodewig: Yeah. It's needed for indentation.

Screenshot of the output without the `+`.

recovery-indentation

docs/en/glossary/glossary.asciidoc
jrodewig: Nor mine. 😄 

It seems obvious I'm gonna need to do some more digging here. Gonna throw a `WIP` label on this PR in the meantime.

docs/en/glossary/glossary.asciidoc
```diff
@@ -552,7 +552,7 @@ endif::cloud-terms[]
 ifdef::elasticsearch-terms[]
 
 [[glossary-recovery]] recovery ::
-The process of updating a shard copy and making it available for queries.
+The process of syncing a shard copy from a source shard. Upon completion, the recovery process makes the shard copy available for queries.
```
jrodewig: Chatted with @dnhatn and settled on this copy.

PR #254, opened 2019-03-22
Add ecs to terminology list

karenzone:
Add Elastic Common Schema (ECS) to the glossary

### DIFF
--- a/docs/en/glossary/glossary.asciidoc
+++ b/docs/en/glossary/glossary.asciidoc
@@ -235,6 +235,13 @@ getting or searching for a document.
 +
 //Source: Elasticsearch
 endif::elasticsearch-terms[]
+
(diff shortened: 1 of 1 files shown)


karenzone:
Moving from a slack thread to here: 

@webmat suggested: 

> "The Elastic Common Schema (ECS) defines a common set of fields, their datatype, and gives guidance on their correct usage. ECS is used to improve uniformity of event data ingested into Elasticsearch."


I like where this is heading. For the glossary, we need to define what ECS _is_, not what it _does_.

webmat:
Here's an attempt at what it actually *is* :-)

> The Elastic Common Schema (ECS) is a document schema for Elasticsearch, for use cases such as logging and metrics. ECS defines a common set of fields, their datatype, and gives guidance on their correct usage. ECS is used to improve uniformity of event data coming from different sources.

PR #266, opened 2019-04-04
[DOCS] Rewrite query def

jrodewig:
Condenses and simplifies the `query` definition from the [Elasticsearch glossary](https://www.elastic.co/guide/en/elasticsearch/reference/current/glossary.html).

Lateral port (sideport?) of elastic/elasticsearch#40746

Resolves elastic/elasticsearch#40643


jrodewig:
Merging without review.

@lcawl already gave a thumbs up to elastic/elasticsearch#40746. This just ports that copy over with a minor change to the link syntax. Verified the link works in my local doc build.

PR #278, opened 2019-04-08
Updates X-Pack and Upgrade Assistant details

lcawl:
This PR clarifies the upgrade process with respect to installing X-Pack and using the Upgrade Assistant.

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -117,15 +117,14 @@ and Beats as soon as possible to take advantage of performance improvements
 and other enhancements.
 
 [discrete]
-=== Upgrading from 5.6 - 6.2
+=== Upgrading from 6.0 - 6.2
 
 Starting in 6.3, the default distributions of Elasticsearch, Logstash, and Kibana
 include {xpack} and a free Basic license that never expires.
 
-You can perform rolling upgrades to {version} from OSS-only clusters running 5.6
-or 6.0-6.2. Basic features are operational once the cluster is fully
-upgraded. If you are already using {xpack}, your settings are preserved when
-you upgrade.
+You can perform rolling upgrades to {version} from OSS-only clusters running
+6.0-6.2. Basic features are operational once the cluster is fully upgraded. If
+you are already using {xpack}, your settings are preserved when you upgrade.
 
 If you are using {xpack} for the first time, you must explicitly enable data
 collection after the upgrade to use monitoring. Set
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -122,7 +122,7 @@ and other enhancements.
 Starting in 6.3, the default distributions of Elasticsearch, Logstash, and Kibana
 include {xpack} and a free Basic license that never expires.
 
-You can perform rolling upgrades to 6.3 from OSS-only clusters running 5.6
+You can perform rolling upgrades to {version} from OSS-only clusters running 5.6
 or 6.0-6.2. Basic features are operational once the cluster is fully
 upgraded. If you are already using {xpack}, your settings are preserved when
 you upgrade.
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -17,9 +17,10 @@ recent 5.6 before upgrading to {version}. {xpack} 5.6
 provides a free Upgrade Assistant that identifies issues you need to address
 before upgrading and simplifies migrating indices that need to be reindexed
 before you upgrade. The Upgrade Assistant is enabled with both Trial and
-Basic licenses. You can install {xpack} solely for the purpose of upgrading. In
-6.3 and later, {xpack} is automatically installed when you install {es}, {kib},
-and Logstash. 
+Basic licenses. You can install {xpack} solely for the purpose of upgrading.
+
+In 6.3 and later, {xpack} is automatically installed when you install {es},
+{kib}, and Logstash. 
 
 Rolling upgrades are supported when upgrading from Elasticsearch 5.6 and
 Elasticsearch 6.0-6.2 to {version}. Upgrading from any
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -6,24 +6,8 @@ each of the products in your Elastic Stack. Beats and Logstash 5.6 are
...
-provides a free Upgrade Assistant that identifies issues you need to address
-before upgrading and simplifies migrating indices that need to be reindexed
-before you upgrade. The Upgrade Assistant is enabled with both Trial and
-Basic licenses. You can install {xpack} solely for the purpose of upgrading. In
-6.3 and later, {xpack} is automatically installed when you install {es}, {kib},
-and Logstash. 
-
-Rolling upgrades are supported when upgrading from Elasticsearch 5.6 and
-Elasticsearch 6.0-6.2 to {version}. Upgrading from any
-version prior to 5.6 requires a full cluster restart.
```
debadair: I don't think it's correct to remove this wholesale. This version covers upgrading *to* 6.7. If you're upgrading from a pre-6.3 version, the information about installing X-Pack still applies, as does using the upgrade guide (which I'm upgrading to mention 6.7). Perhaps we need to split off that last sentence in the second paragraph to make it more obvious? "In
6.3 and later, {xpack} is automatically installed when you install {es}, {kib},
and Logstash. "

lcawl:
You can preview these changes here: https://doc-review-63bf6.firebaseapp.com/x4/upgrading-elastic-stack.html

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -65,15 +49,21 @@ Upgrade Assistant provides information about which {dfeeds} need to be updated.
...
 
-. {ref}/docs-reindex.html[Reindex] or delete any indices created on 2.n. We recommend
-upgrading to the most recent 5.6 and using the {xpack} Reindex Helper to reindex 2.n indices.
-
-. If Kibana and {xpack} are part of your stack, upgrade the internal Kibana
-and {xpack} indices. We recommend using the {xpack} 5.6 Reindex Helper to
-upgrade the internal indices. If you're performing a full cluster restart upgrade
-from an earlier version, you can also <<upgrade-internal-indices,use the
- `_xpack/migration/upgrade` API>> directly to upgrade the
-internal indices after you install Elasticsearch {version}.
```
debadair: This still applies to folks getting to 6.7.

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -65,15 +49,21 @@ Upgrade Assistant provides information about which {dfeeds} need to be updated.
...
+.Upgrade Assistant
+******
+The Upgrade Assistant and migration APIs are enabled with both the Basic and
+Trial licenses. You can install the default distribution of 6.3 to use the
+Upgrade Assistant to prepare to upgrade even if you are upgrading to the OSS
+distribution of {version}.
+******
+
+. Use the Upgrade Assistant to identify any changes you need to make to your
+cluster configuration.
```
debadair: References to 6.3 here don't seem quite right. (And we probably need to tweak the section titled "Upgrading to 6.3" as well, since this is the 6.7 version.

docs/en/install-upgrade/upgrading-stack.asciidoc
lcawl: I moved it down into the body of the upgrade steps to match what you'd done in 7.0, but if you want it repeated still at the top, that's okay by me.

docs/en/install-upgrade/upgrading-stack.asciidoc
lcawl: If you look at the preview, you'll see that's covered in the Upgrade Assistant step.  If it's still not clear, however, let me know.

lcawl: I removed the part about full cluster restarting from an earlier version, since this section assumed you were 5.6 or later, but will see if it's appropriate to add to lower sections.

debadair:
For the "Upgrading to 6.3" section, this might work:

=== Upgrading from versions 5.6-6.2
Starting in 6.3, the default distributions of Elasticsearch, Logstash, and Kibana
include {xpack} and a free Basic license that never expires.
You can perform rolling upgrades to {version} from OSS-only clusters running 5.6
or 6.0-6.2. Basic features are operational once the cluster is fully
upgraded. If you are already using {xpack}, your settings are preserved when
you upgrade.

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -65,15 +49,21 @@ Upgrade Assistant provides information about which {dfeeds} need to be updated.
...
+. Upgrade to the 5.6 or later and use the
+{kibana-ref}/upgrade-assistant.html[{kib} Upgrade Assistant] to
+{ref}/docs-reindex.html[reindex] any indices that are not compatible with {version}.
++
+[role="xpack"]
+.Upgrade Assistant
+******
+The Upgrade Assistant and migration APIs are enabled with both the Basic and
+Trial licenses. You can install the default distribution of 6.3 to use the
+Upgrade Assistant to prepare to upgrade even if you are upgrading to the OSS
```
debadair: This reference to 6.3 is incorrect. You don't need to install 6.3 to get to 6.7. :-)

lcawl: Right, but the default distribution doesn't exist until 6.3. Let's see if I can make that clearer.

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -116,7 +117,8 @@ and Beats as soon as possible to take advantage of performance improvements
 and other enhancements.
 
 [discrete]
-=== Upgrading to 6.3
+=== Upgrading from 5.6 - 6.2
+
```
debadair: The following reference to "rolling upgrades to 6.3" should be "rolling upgrades to {version}"

lcawl:
Thanks for the quick feedback @debadair !

PR #318, opened 2019-04-19
[DOCS] Adds configuration information for ESMS

lcawl:
This PR adds information similar to https://www.elastic.co/guide/en/elastic-stack-overview/master/monitoring-production.html, but instead of sending the monitoring data to a monitoring cluster, it is sent to the Elasticsearch Stack Monitoring Service.

Initially this page contains instructions for using Metricbeat to monitor Kibana and Elasticsearch.  It will be expanded to describe using collectors and exporters and monitoring Logstash.

This PR depends on tagged regions that are implemented in https://github.com/elastic/kibana/pull/37123,  https://github.com/elastic/elasticsearch/pull/42547, https://github.com/elastic/elasticsearch/pull/42601,  and https://github.com/elastic/kibana/pull/37198 to permit content-reuse. It also uses attributes implemented in https://github.com/elastic/docs/pull/902

### DIFF
--- a/docs/en/stack/monitoring/esms.asciidoc
+++ b/docs/en/stack/monitoring/esms.asciidoc
@@ -183,14 +183,7 @@ To use {metricbeat}:
 . Disable the default collection of {kib} monitoring metrics. +
 +
 --
-Add the following setting in the {kib} configuration file (`kibana.yml`): 
-
-[source,yaml]
-----------------------------------
-xpack.monitoring.kibana.collection.enabled: false
-----------------------------------
-
-Leave the `xpack.monitoring.enabled` set to its default value (`true`). 
+include::{kib-repo-dir}/monitoring/monitoring-metricbeat.asciidoc[tag=disable-kibana-collection]
 
 For more information, see 
 {kibana-ref}/monitoring-settings-kb.html[Monitoring settings in {kib}].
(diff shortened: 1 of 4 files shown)


lcawl:
This content can be previewed in https://doc-review-63bf6.firebaseapp.com/x5/esms.html

docs/en/stack/monitoring/esms.asciidoc
```diff
@@ -0,0 +1,208 @@
...
+TIP: If you want to monitor {ls}, you must use collectors and exporters to route
+data from the production cluster to {esms-init}. Otherwise, it is simplest to
+use {metricbeat}. 
+
+To use {metricbeat}:
+
+. If your production cluster is stopped,
+{ref}/starting-elasticsearch.html[start it].
+
+. Configure the following settings on each node in the production cluster:
```
ycombinator: For this step, could we describe to the user _why_ they should configure these settings, kinda like we've done in https://www.elastic.co/guide/en/elasticsearch/reference/master/configuring-metricbeat.html:

> Enable the collection of monitoring data. Set `xpack.monitoring.collection.enabled` to `true` on each node in the production cluster. By default, it is is disabled (`false`).
> Disable the default collection of Elasticsearch monitoring metrics. Set `xpack.monitoring.elasticsearch.collection.enabled` to `false` on each node in the production cluster.

docs/en/stack/monitoring/esms.asciidoc
```diff
@@ -0,0 +1,208 @@
...
+file on any node, you must change or remove it.
+
+For more information about these settings, see
+{ref}/monitoring-settings.html[Monitoring settings in {es}].
+--
+
+. {metricbeat-ref}/metricbeat-installation.html[Install {metricbeat}] on each
+node.
+
+. Optional: Disable the system module in {metricbeat}. +
```
ycombinator: Should we explain why we include this step, especially since it's marked as optional. That way the user can decide whether they should take it or not.

To recap, the reason for this step is to stop Metricbeat from indexing metrics that are currently not used by the ESMS cluster.

docs/en/stack/monitoring/esms.asciidoc
lcawl: I had combined those two settings steps for the sake of brevity, but it's easy enough to just re-use them as-is from the Elasticsearch steps.  I've made that change.

nik9000:
[here](http://stack-docs_318.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/esms.html) is a quick preview of the docs generated by this change. The preview process isn't ready for prime time yet but it did pick this one up.

docs/en/stack/monitoring/esms.asciidoc
lcawl: Thanks, I've added explanatory text in https://github.com/elastic/elasticsearch/pull/42601 and https://github.com/elastic/kibana/pull/37198 and pulled that information into this page too.

ycombinator:
>  It will be expanded to describe using collectors and exporters and monitoring Logstash.

And also Beats (just as a heads up to you 😄).

PR #329, opened 2019-05-08
[DOCS] Updates to the Getting started with security tutorial

lcawl:
This PR updates the following tutorials in the Stack Overview: 
https://www.elastic.co/guide/en/elastic-stack-overview/master/security-getting-started.html
https://www.elastic.co/guide/en/elastic-stack-overview/master/encrypting-internode-communications.html

NOTE: It also changes the step related to looking at the privileges for the `kibana_user` role, since that information was no longer correct.

### DIFF
--- a/docs/en/stack/security/get-started-security.asciidoc
+++ b/docs/en/stack/security/get-started-security.asciidoc
@@ -24,10 +24,16 @@ running.
 . Launch the {kib} web interface by pointing your browser to port 5601. For 
 example, http://127.0.0.1:5601[http://127.0.0.1:5601].
 
-TIP: By default, when you install {stack} products, they apply basic licenses
-with no expiration dates. To view your license in {kib}, go to **Management**
-and click **License Management**. For more information about Elastic license
-levels, see https://www.elastic.co/subscriptions.
+. Verify that you are using a license that includes the role-based access
+control (RBAC) and native authentication {security-features}. To view your
+license in {kib}, go to **Management** and click **License Management**. 
++
+--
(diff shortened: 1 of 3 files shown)


joshbressers:
LGTM

docs/en/stack/security/get-started-security.asciidoc
```diff
@@ -24,11 +24,16 @@ running.
 . Launch the {kib} web interface by pointing your browser to port 5601. For 
 example, http://127.0.0.1:5601[http://127.0.0.1:5601].
 
-[role="xpack"]
-[[get-started-license]]
-=== Install a trial license
-
-include::{docdir}/get-started-trial.asciidoc[]
+. Verify that you are using a license that includes the role-based access
```
skearns64: There are now no license types that don't allow security configuration, so I think we can drop this paragraph? Alternatively, we could use this to ensure folks have the default distribution.

docs/en/stack/security/securing-communications/tutorial-tls-intro.asciidoc
```diff
@@ -27,7 +26,16 @@ described in
 particular, this tutorial provides instructions that work with the `zip` and
 `tar.gz` packages.
 
-. <<get-started-license,Install a trial license>>.
+. Verify that you are using a license that includes the encrypted communications
```
skearns64: Same comment as previous - this paragraph and the following one aren't really necessary if they are using the default distribution, as there are no license types for which TLS isn't supported

PR #351, opened 2019-05-30
Rework file structure to combine hosts and network guides

karenzone:
This PR: 
* changes to file structure to combine network and hosts into one guide
* sets up placeholder text to allow for pre-release linking.   (Note that another step is required before placeholder is published and ready for linking.)


cwurm:
I see `SIEM Monitoring` / `SIEM Solution` / `security monitoring` in some places. I don't think we need the "monitoring/solution/security" piece, just `SIEM` everywhere?

karenzone:
Agreed. I will clean these up as we flesh out the content.

PR #352, opened 2019-05-30
[DOCS] Added Data Frames subsection to ML section.

szabosteve:


docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,29 @@
+[[ml-dataframes]]
+=== Data frames
+++++
+<titleabbrev>Dataframes</titleabbrev>
```
lcawl: This titleabbrev isn't needed unless you want a different title in the table of contents than on the page itself.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,29 @@
+[[ml-dataframes]]
+=== Data frames
```
lcawl: After https://github.com/elastic/docs/pull/902 is merged, this can be changed to:

```suggestion
=== {dataframes-cap}
```

docs/en/stack/ml/dataframes.asciidoc
szabosteve: I deleted it.

docs/en/stack/ml/dataframes.asciidoc
szabosteve: I've already changed it. Maybe that's why the check failed? In my local, the doc build was successful.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
+[[ml-dataframes]]
+=== {dataframes-cap}
+
+{dataframes-cap} feature is available in 7.2 and later.
```
lcawl: I don't think this availability date information is necessary. This page will only exist in 7.2 and later documentation and this functionality be mentioned in the 7.2 release highlights so that should be sufficient.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+
+A _{dataframe}_ is a transformation of a dataset by certain rules defined during
```
lcawl: I think you could consider dropping everything after "rules". i.e. In this high-level overview, I don't think it's necessary to specify *when* you define the rules.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+the creation of the {dataframe}. You can think of it like a spreadsheet or a 
+data table that makes your data ready to be analyzed and organized.
+
+{es} datasets consist of individual documents that have fields and
+values in each field. This architecture makes search easy but on the other hand, 
+makes it hard to run analyses that require reorganized or summarized fields of 
+the dataset. {ml-cap} analyses need clean and transformed data and that is the 
+point where {dataframes} come into play.
+
+To transform the data into a {dataframe}, you need to define a _pivot_. During
```
lcawl: We might want to consider merging these sentences.  For example: "...you need to define a _pivot_, which is a set of features that transform the dataset into a different, more digestible format. Pivoting results..."

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+pivoting, you create a set of features that transform the dataset into a
+different, more digestible format to make calculations on your data. Pivoting
+results in a summary of your dataset (which is the {dataframe} itself).
+
+Defining a pivot consist of two main parts. First, you select one or more fields 
+that your dataset will be grouped by. Principally you can select categorical 
+fields (terms) for grouping. You can also select numerical fields, in this case, 
+the field values will be bucketed using an interval you specify. The calculation
```
lcawl: I find it unclear what "the calculation" is at this point. I would suggest omitting the sentence "The calculation will run...".

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+will run against every bucket that was created this way.
+
+The second step is selecting one or more aggregations to perform calculation over
```
lcawl: Another possible wording for this (per https://github.com/elastic/elasticsearch/pull/42728) is something like this: "The second step is deciding how you want to aggregate the grouped data".

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+scope of the aggregation.
+
+IMPORTANT: In 7.2, you can build {dataframes} on the top of a static dataset. 
+When new data comes into the index, you have to perform the transformation again 
+on the altered data. Using {dataframes} does not require {dfeeds}. 
+{con-dataframes-cap} will be introduced in a later version.
+
+.Example
+
+Put the case that you run a webshop that sells clothes. Every order creates a
```
lcawl:
```suggestion
Imagine that you run a webshop that sells clothes. Every order creates a 
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+on the altered data. Using {dataframes} does not require {dfeeds}. 
+{con-dataframes-cap} will be introduced in a later version.
+
+.Example
+
+Put the case that you run a webshop that sells clothes. Every order creates a 
+document that contains a unique order ID, the name and the category of the 
+ordered product, its price, the ordered quantity, the exact date of the order, 
+and some customer information (name, gender, location, etc). Your dataset 
+contains the documents of all the transactions from last year.
```
lcawl:
```suggestion
contains all the transactions from last year.
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+
+If you want to check the sales in the different categories in your last fiscal year,
+define a {dataframe} that is grouped by the product categories (women's shoes, men's
+clothing, etc.) and the order date histogram with the interval of the last year, 
+then set a sum aggregation on the ordered quantity. The result is a {dataframe} 
+pivot that shows the number of sold items in every product category in the last 
+year.
+
```
lcawl: I think it would be helpful to have a visual representation of the example, even if it's just a made-up set of data in a table format.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+that your dataset will be grouped by. Principally you can select categorical 
+fields (terms) for grouping. You can also select numerical fields, in this case, 
+the field values will be bucketed using an interval you specify. The calculation
+will run against every bucket that was created this way.
+
+The second step is selecting one or more aggregations to perform calculation over
+the dataset. When using aggregations, you practically ask questions about the 
+dataset. There are different types of aggregations, each with its own purpose and 
+output. You can learn more about the supported aggregations and group-by fields 
+here (!add a link!).
```
lcawl: This can likely link to {ref}/data-frame-transform-pivot.html after https://github.com/elastic/elasticsearch/pull/42728/files is merged

docs/en/stack/ml/dataframes.asciidoc
hendrikmuhs: "you need to define a _pivot_" : I think this can be misunderstood. _pivot_ is 1 function of a data frame, it's the only one right now, but there are plans to add more in upcoming versions. Not sure how to phrase that.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
+[[ml-dataframes]]
+=== {dataframes-cap}
+
+{dataframes-cap} feature is available in 7.2 and later.
+
+A _{dataframe}_ is a transformation of a dataset by certain rules defined during
+the creation of the {dataframe}. You can think of it like a spreadsheet or a 
+data table that makes your data ready to be analyzed and organized.
+
+{es} datasets consist of individual documents that have fields and
```
hendrikmuhs: it _depends_ ;-) This sounds a bit to general for me, you can also upload summarized data into elasticsearch or you can summarize data using external tools (e.g. hadoop/spark). Maybe you can say:

A lot of datasets are organized as a stream of events: Each event is a individual document, e.g. a single item purchase, data frame pivots allow you to summarize this data, bringing it into a an organized, more analysis friendly format, e.g. summarizing all purchases of a single customer...

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+The second step is selecting one or more aggregations to perform calculation over
+the dataset. When using aggregations, you practically ask questions about the 
+dataset. There are different types of aggregations, each with its own purpose and 
+output. You can learn more about the supported aggregations and group-by fields 
+here (!add a link!).
+
+As an optional step, it's also possible to add a query to further limit the 
+scope of the aggregation.
+
+IMPORTANT: In 7.2, you can build {dataframes} on the top of a static dataset.
```
hendrikmuhs: is it usual to have roadmap statements like this? I agree it is important to talk about it, but I wonder if this is the right place.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+dataset. There are different types of aggregations, each with its own purpose and 
+output. You can learn more about the supported aggregations and group-by fields 
+here (!add a link!).
+
+As an optional step, it's also possible to add a query to further limit the 
+scope of the aggregation.
+
+IMPORTANT: In 7.2, you can build {dataframes} on the top of a static dataset. 
+When new data comes into the index, you have to perform the transformation again 
+on the altered data. Using {dataframes} does not require {dfeeds}.
```
hendrikmuhs: what does {dfeeds} refer to?

szabosteve: It refers to datafeeds.

docs/en/stack/ml/dataframes.asciidoc
szabosteve: Okay, I'll think about it and rephrase this sentence.

docs/en/stack/ml/dataframes.asciidoc
szabosteve: Thank you for your suggestion, I amended the text accordingly.

docs/en/stack/ml/dataframes.asciidoc
hendrikmuhs: there is no connection between datafeeds and data frames, neither once we have continuous data frames. I wonder if mentioning datafeeds is confusing, I would not mention.

lcawl: I think it's worth mentioning that data frames do not require datafeeds (even if it's a little tip at the end), since this page is alongside all the other ML content where datafeeds *are* required.

pheyos:
Would this documentation be the right place to mention the needed permissions around the whole data frame topic? Asking, because I currently can't find it anywhere and users might think `machine_learning_admin` might be the required role (since data frames are located under ML), which is not the case.

lcawl:
@pheyos We typically add that type of information in the API reference material, but you're right I don't see it in https://www.elastic.co/guide/en/elasticsearch/reference/master/put-data-frame-transform.html for example.  I'll follow-up and add it there and potentially in the (upcoming) configuration example too.

pheyos:
thanks @lcawl !

docs/en/stack/ml/dataframes.asciidoc
szabosteve: I rephrased the sentence to clarify that pivot is a function of a data frame.

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+=== {dataframes-cap}
+
+beta[]
+
+A _{dataframe}_ is a transformation of a dataset by certain rules. You can think 
+of it like a spreadsheet or a data table that makes your data ready to be analyzed 
+and organized.
+
+A lot of {es} datasets are organized as a stream of events: each event is a individual 
+document, e.g. a single item purchase. {dataframe-cap} pivots allow you to summarize
```
lcawl:
```suggestion
Keep in mind that the abbreviations e.g. and i.e. can be difficult to translate. If your work will be translated into another language, consider writing out “for example” and “in other words” instead. 
document, for example a single item purchase. {dataframe-cap} pivots allow you to summarize 
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+
+beta[]
+
+A _{dataframe}_ is a transformation of a dataset by certain rules. You can think 
+of it like a spreadsheet or a data table that makes your data ready to be analyzed 
+and organized.
+
+A lot of {es} datasets are organized as a stream of events: each event is a individual 
+document, e.g. a single item purchase. {dataframe-cap} pivots allow you to summarize 
+this data, bringing it into an organized, more analysis friendly format, e.g.
```
lcawl: Per above, I suggest replacing "e.g."
```suggestion
this data, bringing it into an organized, more analysis friendly format, for example,
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+beta[]
+
+A _{dataframe}_ is a transformation of a dataset by certain rules. You can think 
+of it like a spreadsheet or a data table that makes your data ready to be analyzed 
+and organized.
+
+A lot of {es} datasets are organized as a stream of events: each event is a individual 
+document, e.g. a single item purchase. {dataframe-cap} pivots allow you to summarize 
+this data, bringing it into an organized, more analysis friendly format, e.g. 
+summarizing all purchases of a single customer (see the example below).
```
lcawl:
```suggestion
summarize all the purchases of a single customer (see the example below).
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+results in a summary of your dataset (which is the {dataframe} itself).
+
+Defining a pivot consist of two main parts. First, you select one or more fields 
+that your dataset will be grouped by. Principally you can select categorical 
+fields (terms) for grouping. You can also select numerical fields, in this case, 
+the field values will be bucketed using an interval you specify.
+
+The second step is deciding how you want to aggregate the grouped data. When 
+using aggregations, you practically ask questions about the dataset. There are 
+different types of aggregations, each with its own purpose and output. You can
```
lcawl:
```suggestion
different types of aggregations, each with its own purpose and output. To
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+
+Defining a pivot consist of two main parts. First, you select one or more fields 
+that your dataset will be grouped by. Principally you can select categorical 
+fields (terms) for grouping. You can also select numerical fields, in this case, 
+the field values will be bucketed using an interval you specify.
+
+The second step is deciding how you want to aggregate the grouped data. When 
+using aggregations, you practically ask questions about the dataset. There are 
+different types of aggregations, each with its own purpose and output. You can 
+learn more about the supported aggregations and group-by fields
```
lcawl:
```suggestion
learn more about the supported aggregations and group-by fields, see
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+Defining a pivot consist of two main parts. First, you select one or more fields 
+that your dataset will be grouped by. Principally you can select categorical 
+fields (terms) for grouping. You can also select numerical fields, in this case, 
+the field values will be bucketed using an interval you specify.
+
+The second step is deciding how you want to aggregate the grouped data. When 
+using aggregations, you practically ask questions about the dataset. There are 
+different types of aggregations, each with its own purpose and output. You can 
+learn more about the supported aggregations and group-by fields 
+{ref}/data-frame-transform-pivot.html[here].
```
lcawl: I think it's a good idea to have the title of the destination page in the link, so that people can decide whether they want to follow it (and in case something goes wrong with the link and they need to search for it by its title).
```suggestion
{ref}/data-frame-transform-pivot.html[Pivot resources].
```

docs/en/stack/ml/dataframes.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+
+Imagine that you run a webshop that sells clothes. Every order creates a 
+document that contains a unique order ID, the name and the category of the 
+ordered product, its price, the ordered quantity, the exact date of the order, 
+and some customer information (name, gender, location, etc). Your dataset 
+contains all the transactions from last year.
+
+If you want to check the sales in the different categories in your last fiscal year,
+define a {dataframe} that is grouped by the product categories (women's shoes, men's
+clothing, etc.) and the order date histogram with the interval of the last year,
```
lcawl: I'm not sure what "histogram" adds here, so I suggest omitting it:
```suggestion
clothing, etc.) and the order date with the interval of the last year, 
```

docs/en/stack/ml/dataframes.asciidoc
szabosteve: I added a screenshot from the pivot preview that represents the use case that I used in the example.

PR #360, opened 2019-06-07
Fix confusion on the order parameter

pmusa:
This PR makes the following clearer.
- The order list does not matter. What matters is the `order` parameter.
- Also, it is not clear which order is consulted first.

### DIFF
--- a/docs/en/stack/security/authentication/realm-chains.asciidoc
+++ b/docs/en/stack/security/authentication/realm-chains.asciidoc
@@ -3,8 +3,9 @@
 === Realm chains
 
 <<realms,Realms>> live within a _realm chain_. It is essentially a prioritized list of
-configured realms (typically of various types). The order of the list determines
-the order in which the realms will be consulted. You should make sure each
+configured realms (typically of various types). The `order` setting determines
+the order in which the realms will be consulted. Lower orders will be consulted first.
+You should make sure each
 configured realm has a distinct `order` setting. In the event that two or more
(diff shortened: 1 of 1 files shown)


docs/en/stack/security/authentication/realm-chains.asciidoc
```diff
@@ -3,8 +3,9 @@
 === Realm chains
 
 <<realms,Realms>> live within a _realm chain_. It is essentially a prioritized list of
-configured realms (typically of various types). The order of the list determines
-the order in which the realms will be consulted. You should make sure each
+configured realms (typically of various types). The `order` setting determines
+the order in which the realms will be consulted. Lower orders will be consulted first.
```
lcawl:
```suggestion
Realms are consulted in ascending order (that is to say, the realm with the lowest `order` value is consulted first).
```

pmusa:
I like the suggestion. thanks @lcawl.

docs/en/stack/security/authentication/realm-chains.asciidoc
```diff
@@ -3,8 +3,9 @@
 === Realm chains
 
 <<realms,Realms>> live within a _realm chain_. It is essentially a prioritized list of
-configured realms (typically of various types). The order of the list determines
-the order in which the realms will be consulted. You should make sure each
+configured realms (typically of various types). The `order` setting determines
```
jkakavas:
```suggestion
configured realms (typically of various types).
```

PR #361, opened 2019-06-07
[DOCS] Update ML limitation

lcawl:
This PR updates the following limitation, since it is no longer accurate:

https://www.elastic.co/guide/en/elastic-stack-overview/master/ml-limitations.html#_time_based_index_patterns_are_not_supported

### DIFF
--- a/docs/en/stack/ml/limitations.asciidoc
+++ b/docs/en/stack/ml/limitations.asciidoc
@@ -115,7 +115,7 @@ analyzed.
 
 
 [float]
-=== Using time-based index patterns
+=== Time-based index patterns in {kib}
 //See x-pack-elasticsearch/#1910
 
 If you try to create a {ml} analysis job that uses time-based index patterns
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/limitations.asciidoc
```diff
@@ -115,13 +115,13 @@ analyzed.
...
 [float]
-=== Time-based index patterns are not supported
+=== Time-based index patterns in {kib}
 //See x-pack-elasticsearch/#1910
 
-It is not possible to create a {ml} analysis job that uses time-based
-index patterns, for example `[logstash-]YYYY.MM.DD`.
-This applies to the single metric or multi metric job creation wizards in {kib}.
-
+If you try to create a {ml} analysis job that uses time-based index patterns
```
peteharverson: I think this whole section can be dropped. I don't think it is even possible to create an index pattern like `[logstash-]YYYY.MM.DD` in the Kibana index patterns page. The docs only refer to using wlildcarded index patterns - see https://www.elastic.co/guide/en/kibana/current/tutorial-define-index.html.

lcawl: Thanks, I've removed that section entirely.

PR #371, opened 2019-06-13
 [DOCS] Adds data frames section under limitations

szabosteve:
This PR creates a page that is used in the Stack Overview. It provides information about the known issues/limitations of the data frame feature. Related issue: /elastic/ml-team/issues/167


docs/en/stack/data-frames/index.asciidoc
```diff
@@ -1,3 +1,4 @@
 include::dataframes.asciidoc[]
 include::ecommerce-example.asciidoc[]
-include::api-quickref.asciidoc[]
\ No newline at end of file
+include::api-quickref.asciidoc[]
+include::dataframe-limitations.asciidoc[]
```
lcawl: There's actually a "Limitations" section in the Stack Overview: https://www.elastic.co/guide/en/elastic-stack-overview/master/limitations.html
I think until we rethink the layout of this book, we should follow the existing pattern and put the "Data frames" page there too.

docs/en/stack/data-frames/dataframe-limitations.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+beta[]
+
+The following limitations and known problems apply to the 7.2 release of 
+the Elastic {dataframe} feature:
+
+[float]
+[[df-limitations-wizard]]
+=== {dataframe-cap} wizard limitation
+
+The {dataframe} wizard in {kib} cannot display the value of a filed selected by
```
lcawl:
```suggestion
The {dataframe} wizard in {kib} cannot display the value of a field selected by 
```

docs/en/stack/data-frames/dataframe-limitations.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+The following limitations and known problems apply to the 7.2 release of 
+the Elastic {dataframe} feature:
+
+[float]
+[[df-limitations-wizard]]
+=== {dataframe-cap} wizard limitation
+
+The {dataframe} wizard in {kib} cannot display the value of a filed selected by 
+using the source index table on the right side if the selected field contains an 
+array.
```
lcawl: This is quite a long sentence.  Could we make it easier to understand by doing something like this?:

If you have fields that contain arrays, you cannot select these fields in the source index table in the Kibana data frame wizard.

lcawl: Should we add something like this too?:

> You can, however, use those fields when you group and aggregate your data in the wizard.

docs/en/stack/data-frames/dataframe-limitations.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+[[df-limitations-ccs]]
+=== {ccs-cap} limitation
+
+{ccs-cap} is not supported in 7.2 for {dataframe-transforms}.
+
+[float]
+[[df-limitations-kibana]]
+=== Displaying {dataframe-transforms} in {kib} limitation
+
+{kib} displays {dataframe-transforms} up to 100.
```
lcawl:
```suggestion
The {kib} *Machine Learning* > *Data Frames* page lists up to 100 {dataframe-transforms}.
```

lcawl: I think it's likely worth mentioning that you can work-around this limitation by calling the get data frame transforms API with the `size` parameter.

docs/en/stack/data-frames/dataframe-limitations.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+[float]
+[[df-limitations-dateformat]]
+=== Date histogram limitation
+
+When a date histogram is used as a group by field, the defined interval and the 
+used time format must have the same time fidelity, otherwise it might cause issues 
+in the {dataframe}.
+
+For example, if you set the interval of the date histogram to one minute (1m), 
+then make sure that the time format is `yyyy-MM-dd HH:mm` instead of
```
lcawl: It's not obvious to me where we're telling them to set the format of the date.  Do they need to specify it in the "format" field in the date_histogram aggregation (per https://www.elastic.co/guide/en/elasticsearch/reference/master/search-aggregations-bucket-datehistogram-aggregation.html)?  If so, I don't think they can accomplish this within the Kibana wizard.

docs/en/stack/data-frames/dataframe-limitations.asciidoc
```diff
@@ -0,0 +1,48 @@
...
+`yyyy-MM-dd HH:00`.
+
+[float]
+[[df-limitations-aggresponse]]
+=== Strings in aggregation responses
+
+If an aggregation response contains a string (for example, due to a typo in the 
+field name that results in a "NaN" response), the {dataframe} will fail to index 
+the value without warning or causing an error in the {es} logs. However, it results 
+in an empty destination index.
```
lcawl: This sounds like a bug that we are hopefully planning to fix. If so, perhaps we can include a link to the issue where that's tracked?

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+
+[float]
+[[df-dateformat-limitations]]
+=== Date histogram limitation
+
+When you use the {ref}/data-frame-apis.html[{dataframe-transforms} API] and a date 
+histogram is used as a group by field, the defined interval and the used time format 
+must have the same time fidelity, otherwise it might cause issues in the {dataframe}.
+
+For example, if you set the interval of the date histogram to one minute (1m),
```
lcawl:
```suggestion
For example, if you set the `calendar_interval` of the date histogram to one minute (`1m`), 
```

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+[float]
+[[df-dateformat-limitations]]
+=== Date histogram limitation
+
+When you use the {ref}/data-frame-apis.html[{dataframe-transforms} API] and a date 
+histogram is used as a group by field, the defined interval and the used time format 
+must have the same time fidelity, otherwise it might cause issues in the {dataframe}.
+
+For example, if you set the interval of the date histogram to one minute (1m), 
+then make sure that the time format is `yyyy-MM-dd HH:mm` instead of
```
lcawl:
```suggestion
then make sure that the `format` is `yyyy-MM-dd HH:mm` instead of 
```

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+The {kib} *Machine Learning* > *Data Frames* page lists up to 100 {dataframe-transforms}.
+You can work-around this limitation by calling the 
+{ref}/get-data-frame-transform.html[GET {dataframe-transforms} API] 
+with the `size` parameter.
+
+[float]
+[[df-dateformat-limitations]]
+=== Date histogram limitation
+
+When you use the {ref}/data-frame-apis.html[{dataframe-transforms} API] and a date
```
lcawl:
```suggestion
If you use a {ref}/search-aggregations-bucket-datehistogram-aggregation.html[date
```

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+You can work-around this limitation by calling the 
+{ref}/get-data-frame-transform.html[GET {dataframe-transforms} API] 
+with the `size` parameter.
+
+[float]
+[[df-dateformat-limitations]]
+=== Date histogram limitation
+
+When you use the {ref}/data-frame-apis.html[{dataframe-transforms} API] and a date 
+histogram is used as a group by field, the defined interval and the used time format
```
lcawl:
```suggestion
histogram] in the `group_by` object in the create or preview {dataframe-transform} APIs, the defined interval and time format 
```

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+[float]
+[[df-aggresponse-limitations]]
+=== Strings in aggregation responses
+
+If an aggregation response contains a string (for example, due to a typo in the 
+field name that results in a "NaN" response), the {dataframe} will fail to index 
+the value without warning or causing an error in the {es} logs. However, it results 
+in an empty destination index.
+
+This is a https://github.com/elastic/elasticsearch/issues/43194[known issue] that
```
lcawl: I feel a little uncomfortable promising that an issue will be fixed--I'm sure it will be but maybe it's better to just say something like this:
```suggestion
See https://github.com/elastic/elasticsearch/issues/43194[Issue #43194].
```

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+{ref}/get-data-frame-transform.html[GET {dataframe-transforms} API] 
+with the `size` parameter.
+
+[float]
+[[df-dateformat-limitations]]
+=== Date histogram limitation
+
+If you use a {ref}/search-aggregations-bucket-datehistogram-aggregation.html[date 
+histogram] in the `group_by` object in the create or preview {dataframe-transform} 
+APIs, the defined interval and time format must have the same time fidelity,
```
lcawl: Not a hard requirement, but can shorten this sentence as follows:
```suggestion
APIs, the defined interval and time format must have the same time fidelity.
```

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+with the `size` parameter.
+
+[float]
+[[df-dateformat-limitations]]
+=== Date histogram limitation
+
+If you use a {ref}/search-aggregations-bucket-datehistogram-aggregation.html[date 
+histogram] in the `group_by` object in the create or preview {dataframe-transform} 
+APIs, the defined interval and time format must have the same time fidelity, 
+otherwise it might cause issues in the {dataframe}.
```
lcawl:
```suggestion
Otherwise, it might cause issues in the {dataframe}.
```

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+
+For example, if you set the `calendar_interval` of the date histogram to one minute 
+(`1m`), then make sure that the `format` is `yyyy-MM-dd HH:mm` instead of 
+`yyyy-MM-dd HH:00`.
+
+[float]
+[[df-aggresponse-limitations]]
+=== Strings in aggregation responses
```
sophiec20: This particular issue has been fixed.. however there is a window for some similar silent failures... perhaps change to something along the lines of  ..


`Aggregation responses may be incompatible with destination index mappings`

`Data frames use composite aggregations in order to transform data. In some cases, composite aggregations may return responses which are not compatible with the mappings set for the destination index. For example "NaN", "Infinity" or possibly a numeric overflow. Where possible, a null response has been substituted. Please check elasticsearch logs if you think this may have occurred. As a workaround, using the API, you may define custom destination index mappings prior to starting the data frame transform.`

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+
+If you have fields that contain arrays, you cannot select these fields in the source 
+index table in the {kib} {dataframe} wizard. You can, however, use those fields 
+when you group and aggregate your data in the wizard.
+
+[float]
+[[df-ccs-limitations]]
+=== {ccs-cap} limitation
+
+{ccs-cap} is not supported in 7.2 for {dataframe-transforms}.
```
sophiec20: This is by design ... not sure if we need this documented as a limitations, as it should be documented against the `transform_id`.

- only allow lowercase alphanumeric (a-z and 0-9), hyphens or underscores
- must start and end with alphanumeric
- maximum length of 64 characters

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+== {dataframe-cap} limitations
+
+beta[]
+
+The following limitations and known problems apply to the 7.2 release of 
+the Elastic {dataframe} feature:
+
+[float]
+[[df-wizard-limitations]]
+=== {dataframe-cap} wizard limitation
```
sophiec20: Data frames does not (yet) support fields containing arrays - in the UI or the API.

If you try to create one, the UI will fail to show the source index table (which is not the main point of this limitation)

docs/en/stack/data-frames/limitations.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+
+[float]
+[[df-ccs-limitations]]
+=== {ccs-cap} limitation
+
+{ccs-cap} is not supported in 7.2 for {dataframe-transforms}.
+
+[float]
+[[df-kibana-limitations]]
+=== Displaying {dataframe-transforms} in {kib} limitation
```
sophiec20: Change title to `{dataframe-transforms} in {kib} will only display up to 100 {dataframe-transforms}`

sophiec20: (note - this will increase to 1000 with 7.2.1)

docs/en/stack/data-frames/limitations.asciidoc
lcawl: I was able to create them (or so I thought) while playing with the eCommerce data. Are they unsupported nonetheless?

PR #372, opened 2019-06-14
[DOCS] Adds data frame eCommerce example

lcawl:
This PR adds a page in the Stack Overview which describes how to use data frame transforms on the Kibana sample eCommerce data.


lcawl:
You can preview this new page here: https://doc-review-63bf6.firebaseapp.com/x/ecommerce-dataframes.html

docs/en/stack/data-frames/ecommerce-example.asciidoc
```diff
@@ -0,0 +1,245 @@
...
+[source,js]
+--------------------------------------------------
+POST _data_frame/transforms/_preview
+{
+  "source": {
+    "index": "kibana_sample_data_ecommerce",
+    "query": {
+      "query_string": {
+        "query": "currency : EUR",
+        "default_operator": "AND"
```
benwtrent: It would be good to make this an actual filter
```
"query": {
   "bool": {
      "filter": {"term": {"currency": "EUR"}}
   }
}
```
Same goes for other queries used.

docs/en/stack/data-frames/ecommerce-example.asciidoc
```diff
@@ -0,0 +1,245 @@
...
+--------------------------------------------------
+// CONSOLE
+// TEST[skip:setup kibana sample data]
+--
+
+. Start the {dataframe-transform}.
++
+--
+
+TIP: A {dataframe-transform} increases search and indexing load on your cluster
```
benwtrent: Might be nice to add something like `Even though resource utilization is automatically adjusted based on cluster load, a {dataframe-transform}...`

We adjust the pagination size of the aggregations whenever we hit a OOM circuit breaker so that we don't break due to cluster load.

docs/en/stack/data-frames/ecommerce-example.asciidoc
lcawl: Thanks @benwtrent I've updated the preview and create data frame transform API examples to use that filter.  I've also updated the query-related tip to recommend that folks use saved searches for more complex queries in Kibana.

PR #389, opened 2019-06-26
[DOCS] Creates a data frame examples page

szabosteve:
This PR creates a data frame examples page under data-frames and indicates it in the index file.


docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+=== Finding suspicious client IPs by using scripted metrics
+
+With {dataframe-transforms}, you can use 
+{ref}/search-aggregations-metrics-scripted-metric-aggregation.html[scripted 
+metric aggregations] on your data. These aggregations are flexible and make 
+it possible to perform very complex processing. Let's use scripted metrics to 
+identify suspicious client IPs in the web log 
+{kibana-ref}/add-sample-data.html[{kib} sample dataset].
+
+The example below transforms web log data into an entity-centric index where the
```
lcawl: In my opinion, this introductory sentence doesn't give me too much context of what I'm about to see in the example. If we can summarize what the transform is going to do, I think that's more helpful.  Something like:

For example, let's transform the data such that the new index contains the sum of bytes, the number of distinct URLs, and the number of distinct geographic destinations for each client IP. We'll also use a scripted field to count the specific types of HTTP responses that each client IP receives.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+    ...
+  }
+----------------------------------  
+// CONSOLE
+// TEST[skip:SETUP]
+
+This {dataframe-transform} makes it easier to answer questions such as:
+
+* Which client IPs are transferring the most amounts of data?
+  (Based on the `"bytes_sum": { "sum": { "field": "bytes" }},` aggregation.)
```
lcawl: I think if we set it up properly in the introduction, we don't need to have the "Based on the ..." phrases here.

docs/en/stack/data-frames/dataframe-examples.asciidoc
szabosteve: You're right. That's why I added the "Based on..." phrases after the questions at the bottom of the doc, but it is a much better solution. I've tried to keep the example relatively short in terms of prose because the API request and the response are fairly long. However, my intention of keeping it short would hurt the overall lucidity of the example. I amended accordingly.

docs/en/stack/data-frames/dataframe-examples.asciidoc
szabosteve: As above.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+<1> Two `group_by` fields have been selected. This means the {dataframe} will 
+contain a unique row per `user` and `customer_id` combination. Within this 
+dataset both these fields are unique. By including both in the {dataframe} it 
+gives more context to the final results.
+
+NOTE: In the example above, condensed JSON formatting has been used for easier 
+readability of the pivot object.
+
+The API returns the following response (note that this example contains the 
+response partially and shows only the first document object):
```
lcawl: I don't think it's necessary to explain that this is just a subset of the results, since `...` is used pretty widely in our examples.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+
+<1> Two `group_by` fields have been selected. This means the {dataframe} will 
+contain a unique row per `user` and `customer_id` combination. Within this 
+dataset both these fields are unique. By including both in the {dataframe} it 
+gives more context to the final results.
+
+NOTE: In the example above, condensed JSON formatting has been used for easier 
+readability of the pivot object.
+
+The API returns the following response (note that this example contains the
```
lcawl: When I first read this I missed that the API was the preview DF API instead of the create DF API. It might be nice to call attention to this and note its usefulness for situations like these where you're playing with DF options. For example, something like this:
 
```suggestion
The preview {dataframe-transforms} API enables you to see the layout of the {dataframe} in advance, populated with some sample values. For example:
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+----------------------------------
+{
+  "preview" : [
+    {
+      "total_unique_products" : {
+        "avg" : 2.0
+      },
+      "customer_id" : "10",
+      "user" : "recip",
+      "order_id" : {
```
lcawl: I think this is the wrong output, since the preview API listed above doesn't include order_id or taxful_total_price, etc. Here's what I get when I run the API on 7.2:

> {
  "preview" : [
    {
      "total_order_amt" : 3946.9765625,
      "order_count" : 59.0,
      "total_unique_products" : 116.0,
      "avg_unique_products_per_order" : 2.0,
      "customer_id" : "10",
      "user" : "recip",
      "avg_amt_per_order" : 66.89790783898304
    },
    ... 
]
}

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+        }
+      }
+    },
+    ...
+  }
+----------------------------------
+// CONSOLE
+// TEST[skip:SETUP]
+
+In the above example we saw how to transform order data into a customer centric
```
lcawl: This "In the above..." sentence is mostly a repeat of what appears in a subsequent paragraph, so I'd remove it here.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+      }
+    },
+    ...
+  }
+----------------------------------
+// CONSOLE
+// TEST[skip:SETUP]
+
+In the above example we saw how to transform order data into a customer centric 
+index. Doing this makes it easier to answer questions such as:
```
lcawl:
```suggestion
This {dataframe} makes it easier to answer questions such as:
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+{dataframes} allow us to persist this data as a customer centric index. This 
+enables us to analyze data at scale and gives more flexibility to explore and 
+navigate data from a customer centric perspective. In some cases, it can even 
+make creating visualizations much simpler.
+
+[float]
+[[example-airline]]
+=== Finding air carriers with the most delays
+
+We use the Flights sample dataset to find out which air carrier delayed the most.
```
lcawl: I suggest that we introduce this example the same way as we did the first one:
```suggestion
In this example, we use the Flights sample dataset to find out which air carrier delayed the most.
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+First, we filter the source data such that excludes all the cancelled flights by
```
lcawl:
```suggestion
First, we filter the source data such that it excludes all the cancelled flights by 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+navigate data from a customer centric perspective. In some cases, it can even 
+make creating visualizations much simpler.
+
+[float]
+[[example-airline]]
+=== Finding air carriers with the most delays
+
+We use the Flights sample dataset to find out which air carrier delayed the most.
+First, we filter the source data such that excludes all the cancelled flights by 
+using a query filter, then transform the data to contain the distinct number of
```
lcawl:
```suggestion
using a query filter. Then we transform the data to contain the distinct number of 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+    }
+  }
+}
+----------------------------------
+// CONSOLE
+// TEST[skip:SETUP]
+
+<1> Filter the source data to select only flights that were not cancelled.
+<2> The data is grouped by the `Carrier` field which contains the airline name.
+<3> This `bucket_script` performs calculations on the results that returned by
```
lcawl:
```suggestion
<3> This `bucket_script` performs calculations on the results that are returned by 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+  }
+}
+----------------------------------
+// CONSOLE
+// TEST[skip:SETUP]
+
+<1> Filter the source data to select only flights that were not cancelled.
+<2> The data is grouped by the `Carrier` field which contains the airline name.
+<3> This `bucket_script` performs calculations on the results that returned by 
+the aggregation, in this particular example to calculate what percentage of
```
lcawl:
```suggestion
the aggregation. In this particular example, it calculates what percentage of 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+{
+  "preview" : [
+    {
+      "carrier" : "ES-Air",
+      "flights_count" : 2802.0,
+      "flight_mins_total" : 1436927.5130677223,
+      "delay_time_percentage" : 9.335543983955839,
+      "delay_mins_total" : 134145.0
+    },
+    {
```
lcawl: I think you can just show one carrier followed by ..., same as in the first example.

docs/en/stack/data-frames/dataframe-examples.asciidoc
lcawl: P.S. I love that these examples use the Kibana sample data, which makes it super easy for me (and our users) to test!

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+// CONSOLE
+// TEST[skip:SETUP]
+
+<1> Filter the source data to select only flights that were not cancelled.
+<2> The data is grouped by the `Carrier` field which contains the airline name.
+<3> This `bucket_script` performs calculations on the results that returned by 
+the aggregation, in this particular example to calculate what percentage of 
+travel time was taken up by delays.
+
+The API returns the following response:
```
lcawl: I think we can alter this sentence in similar ways to my suggestion for the first example.  Something along the lines of:

> The {dataframe} preview shows you that the new index would contain data like this for each carrier:

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+      "delay_time_percentage" : 9.544959286661593,
+      "delay_mins_total" : 143520.0
+    }
+  ]
+}
+----------------------------------
+// CONSOLE
+// TEST[skip:SETUP]
+
+The example above transforms flight data into a entity centric index for
```
lcawl: Unless this first sentence would be meaningful to data scientists, I'd omit it since in my opinion it's unnecessary.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+      "delay_mins_total" : 143520.0
+    }
+  ]
+}
+----------------------------------
+// CONSOLE
+// TEST[skip:SETUP]
+
+The example above transforms flight data into a entity centric index for 
+flight carriers. This makes it easier to answer questions such as:
```
lcawl:
```suggestion
This {dataframe} makes it easier to answer questions such as:
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+
+* Which air carrier has the most delays as a percentage of flight time?
+
+NOTE: Please note that this data is fictional and does not reflect actual delays
```
lcawl: "Please note" is unnecessary inside a "NOTE:"

```suggestion
NOTE: This data is fictional and does not reflect actual delays 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
lcawl: I think since we link to the scripted_metric reference material in the third example, we should link to https://www.elastic.co/guide/en/elasticsearch/reference/master/search-aggregations-pipeline-bucket-script-aggregation.html here too.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+{ref}/search-aggregations-metrics-scripted-metric-aggregation.html[scripted 
+metric aggregations] on your data. These aggregations are flexible and make 
+it possible to perform very complex processing. Let's use scripted metrics to 
+identify suspicious client IPs in the web log sample dataset.
+
+We transform the data such that the new index contains the sum of bytes and the 
+number of distinct URLs, agents, incoming requests by location, and geographic 
+destinations for each client IP. We also use a scripted field to count the 
+specific types of HTTP responses that each client IP receives. Ultimately, the 
+example below transforms web log data into an entity-centric index where the
```
lcawl: Here "entity-centric" has a hyphen and in other cases on this page it does not.  We should be consistent one way or the other.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+
+<1> This range query limits the transform to documents that are within the 
+last 30 days at the point in time the {dataframe-transform} is started.
+<2> The data is grouped by the `clientip` field. 
+<3> This `scripted_metric` performs a distributed operation on the web log data 
+to count specific types of HTTP responses (error, success, and other).
+<4> This `bucket_script` calculates the duration of the `clientip` access based 
+on the results of the aggregation.
+
+The API returns the following response (note that this example contains the
```
lcawl: Per suggestions for  first two examples, I recommend changing this introductory sentence to something like this:

> The {dataframe} preview shows you that the new index would contain data like this for each client IP:

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,362 @@
...
+      },
+      "url_dc" : 11.0
+    },
+    ...
+  }
+----------------------------------  
+// CONSOLE
+// TEST[skip:SETUP]
+
+This {dataframe-transform} makes it easier to answer questions such as:
```
lcawl: >  This {dataframe-transform} ....

I used `{dataframe}` in my earlier suggestions (since I thought the transform is was just the job that generates a data frame) but if I've got that wrong please just use the same term in all three places.

docs/en/stack/data-frames/dataframe-examples.asciidoc
szabosteve: :o Huge thanks for catching this. I played with the API and eventually copied the wrong output.
Lesson learned!

docs/en/stack/data-frames/dataframe-examples.asciidoc
szabosteve: I added the link to the first paragraph of the example as I did the same in the third one.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,333 @@
...
+[testenv="basic"]
+[[dataframe-examples]]
+== {dataframe-cap} examples
+++++
+<titleabbrev>Examples</titleabbrev>
+++++
+
+beta[]
+
+This page provides examples of how to use {dataframe-transforms} to derive useful
```
lcawl:
```suggestion
These examples demonstrate how to use {dataframe-transforms} to derive useful 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,333 @@
...
+* <<example-airline>>
+* <<example-clientips>>
+
+[float]
+[[example-best-customers]]
+=== Finding your best customers
+
+In this example, we use the eCommerce orders sample dataset to find the customers 
+who spent the most in our hypothetical webshop. Let's transform the data such 
+that the destination index contains the number of orders, the total price of
```
lcawl: I know I suggested listing these here, but if you find it's too wordy, maybe a solution would be to have a subset, e.g.:

> Let's create/preview?/consider?/look at? a data frame that contains information such as the ..." (and only list a few interesting items).

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,333 @@
...
+enables us to analyze data at scale and gives more flexibility to explore and 
+navigate data from a customer centric perspective. In some cases, it can even 
+make creating visualizations much simpler.
+
+[float]
+[[example-airline]]
+=== Finding air carriers with the most delays
+
+In this example, we use the Flights sample dataset to find out which air carrier 
+had the most delays.First, we filter the source data such that it excludes all
```
lcawl:
```suggestion
had the most delays. First, we filter the source data such that it excludes all 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,333 @@
...
+// CONSOLE
+// TEST[skip:SETUP]
+
+<1> Filter the source data to select only flights that were not cancelled.
+<2> The data is grouped by the `Carrier` field which contains the airline name.
+<3> This `bucket_script` performs calculations on the results that are returned 
+by the aggregation. In this particular example, it calculates what percentage of 
+travel time was taken up by delays.
+
+The preview {dataframe-transforms} API shows you that the new index would contain
```
lcawl: Not a necessity, but it seems to me you can drop part of this:
```suggestion
The preview shows you that the new index would contain 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,333 @@
...
+
+<1> This range query limits the transform to documents that are within the 
+last 30 days at the point in time the {dataframe-transform} is started.
+<2> The data is grouped by the `clientip` field. 
+<3> This `scripted_metric` performs a distributed operation on the web log data 
+to count specific types of HTTP responses (error, success, and other).
+<4> This `bucket_script` calculates the duration of the `clientip` access based 
+on the results of the aggregation.
+
+The preview {dataframe-transforms} API shows you that the new index would contain
```
lcawl: Ditto:
```suggestion
The preview shows you that the new index would contain 
```

docs/en/stack/data-frames/dataframe-examples.asciidoc
szabosteve: Maybe it is a bit longer this way, but I think it is also more exact. I would leave as is.

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,333 @@
...
+      }
+    }
+  }
+}
+----------------------------------
+// CONSOLE
+// TEST[skip:SETUP]
+
+<1> This range query limits the transform to documents that are within the 
+last 30 days at the point in time the {dataframe-transform} is started.
```
sophiec20: The range query is evaluated as each checkpoint is processed. For batch data frames, there is only ever 1 checkpoint so this statement is true. However I think we should make this statement also cover continuous data frames which are coming soon. 

Perhaps something along the lines of...
`This range query limits the transform to documents that are within the 
last 30 days at the point in time the {dataframe-transform} checkpoint is processed. For batch data frames this occurs once.`

docs/en/stack/data-frames/dataframe-examples.asciidoc
```diff
@@ -0,0 +1,333 @@
...
+and the total amount of ordered products for each customer.
+
+[source,js]
+----------------------------------
+POST _data_frame/transforms/_preview
+{
+  "source": {
+    "index": "kibana_sample_data_ecommerce"
+  },
+  "dest" : {
```
sophiec20: Please add `dest` as a `<numeric-ref>` for all examples - this is the destination index for the data frame and is ignored by `_preview`.

PR #395, opened 2019-07-01
Document "manage_api_key" cluster privilege

tvernum:
Relates: elastic/elasticsearch#43728


docs/en/stack/security/authorization/privileges.asciidoc
```diff
@@ -22,6 +22,14 @@ This includes snapshotting, updating settings, and rerouting. It also includes
 obtaining snapshot and restore status. This privilege does not include the 
 ability to manage security.
 
+`manage_api_key`::
+All security-related operations on {es} API Keys, including
```
lcawl: Unless there's a reason why "Keys" is capitalized here, I think it should be lowercase:
```suggestion
All security-related operations on {es} API keys, including 
```

PR #404, opened 2019-07-09
Monitoring Logstash and Beats with Metricbeat

lcawl:
Related to https://github.com/elastic/logstash/issues/10920 and https://github.com/elastic/beats/issues/12647
This PR updates the Stack Overview's monitoring summary to reflect that Logstash and Beats can be monitored via Metricbeat.


cachedout:
@lcawl Anything that @ycombinator or I can do to help on this one?

lcawl:
@cachedout I've commented out the architectural diagrams for now, since they need to be updated to include Logstash and Beats and I think they could bear an overhaul in general.  If think you've expressed opinions about how to improve those diagrams in the past, so any input is welcome.

lcawl:
The changed pages can be previewed here:
http://stack-docs_404.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/how-monitoring-works.html
http://stack-docs_404.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/monitoring-production.html
http://stack-docs_404.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/esms.html

cachedout:
Thanks! A few comments below:

http://stack-docs_404.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/how-monitoring-works.html

> You can use Metricbeat to collect and ship data about Elasticsearch, Kibana, Logstash, and Beats rather than routing it through Elasticsearch.

This feels a little confusing. I think that by "routing it through Elasticsearch", we mean, "routing it through the _production_ Elasticsearch cluster to the dedicated monitoring cluster". Since both clusters are Elasticsearch, perhaps we can disambiguate this? 

> If you use Kibana to visualize data and administer the cluster, you might want to create a dedicated Kibana instance for monitoring, rather than using a single Kibana instance to access both your production cluster and monitoring cluster.

This feels like a duplicate of the last sentence that's in the "Tip" paragraph, we we write: "For the same reason, we also recommend to use a separate Kibana instance that is connected to the separate monitoring cluster."

http://stack-docs_404.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/monitoring-production.html

> By default, monitoring agents store data in the cluster where they’re running. 

I am having some trouble parsing the intended meaning of this sentence. Perhaps this was the pre-Metricbeat world? If we're updating these docs to include information about Metricbeat, I think we need to standardize on language for "internal" (or we often call "native" monitoring, though it's also referred to in the docs as "exporters" and "collectors") and "external", or what we now call Beats monitoring. In the sentence above, it is unclear which of those forms we are referring to by the word "agent".

lcawl:
> ... I think that by "routing it through Elasticsearch", we mean, "routing it through the production Elasticsearch cluster to the dedicated monitoring cluster". Since both clusters are Elasticsearch, perhaps we can disambiguate this?

Yes, that was the intention. I've updated that text.

> If you use Kibana to visualize data and administer the cluster... This feels like a duplicate of the last sentence that's in the "Tip" paragraph

Agreed. I've removed that paragraph and diagram.

> By default, monitoring agents store data in the cluster where they’re running... I am having some trouble parsing the intended meaning of this sentence. Perhaps this was the pre-Metricbeat world?

Yes, that's definitely from the world of collectors. I agree it's confusing so I've omitted it.

Thanks for the feedback!

PR #431, opened 2019-07-17
Adds data frame transform troubleshooting info

lcawl:
This PR adds information about where to gather diagnostic data for data frame transforms.


docs/en/stack/data-frames/troubleshooting.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+If you encounter problems with your {dataframe-transforms}, you can gather more
+information from the following files and APIs:
+
+* Lightweight audit messages are stored in `.data-frame-notifications-*`. Search
+by your `transform_id`.
+* The
+{ref}/get-data-frame-transform-stats.html[get {dataframe-transform} statistics API] 
+provides information about the transform status and failures.
+* If the {dataframe-transform} exists as a task, you can use the
```
sophiec20: it should exist as a task if it is in either the failed or started state.

docs/en/stack/data-frames/troubleshooting.asciidoc
```diff
@@ -0,0 +1,29 @@
+[[dataframe-troubleshooting]]
+== Troubleshooting {dataframe-transforms}
+[subs="attributes"]
+++++
+<titleabbrev>Troubleshooting</titleabbrev>
+++++
+
+Use the information in this section to troubleshoot common problems and find
```
sophiec20: Not sure we have enough in here right now to call it an FAQ.

docs/en/stack/data-frames/troubleshooting.asciidoc
lcawl: Thanks, added that sentence.

PR #438, opened 2019-07-24
 [DOCS] Makes DFT overview page more readable

szabosteve:
This PR increases the readability of the data frame transforms overview page by adding section headings to the text.

Addresses https://github.com/elastic/ml-team/issues/187#issuecomment-513263269

### DIFF
--- a/docs/en/stack/data-frames/overview.asciidoc
+++ b/docs/en/stack/data-frames/overview.asciidoc
@@ -7,6 +7,10 @@
 
 beta[]
 
+[discrete]
+[[ml-what-is-a-dataframe]]
+== What is a {dataframe}?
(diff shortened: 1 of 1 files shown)


lcawl:
I'm not sure these are required now that the content has been re-organized. If they are, my two cents are as follows:
1) I don't think it's useful to have the ml-what-is-a-dataframe section. That should be obvious from the name the page (and if not, we should consider renaming the page).
2) In my opinion, the second section is not about pivoting, it's about creating a data frame transform. So if we need that section, I'd suggest a title along those lines.

szabosteve:
Okay, thank you for sharing your opinion. I close this PR.

PR #453, opened 2019-08-02
fix typo in command-line for certificate generation

faec:
This fixes two issues with the example command to generate a node certificate:

- it adds a backslash that was missing on one of the lines, which prevented the subsequent flag from being read
- it removes the blank space between the end-of-line backslashes and the line markings (e.g. `"\ <2>"` becomes `"\<2>"`), which added invisible spaces to the command when copied-and-pasted, resulting in most of the lines being ignored even though it looked right on the terminal

@lcawl, I don't know much about our asciidoc practices / build system, and my only concern is that the second item (removing the spaces) might keep it from recognizing the footnote symbols. If so I imagine you can advise me how to do it correctly :-)

### DIFF
--- a/docs/en/stack/security/securing-communications/tutorial-tls-certificates.asciidoc
+++ b/docs/en/stack/security/securing-communications/tutorial-tls-certificates.asciidoc
@@ -48,10 +48,10 @@ Run the following command:
 ["source","sh",subs="attributes,callouts"]
 ----------------------------------------------------------------------
 ./bin/elasticsearch-certutil cert \
-  --ca elastic-stack-ca.p12 \<1>
-  --dns localhost \<2>
-  --ip 127.0.0.1,::1 \<3>
-  --out config/certs/node-1.p12 <4>
+  --ca elastic-stack-ca.p12 \//<1>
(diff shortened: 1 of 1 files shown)


lcawl:
It seems to work by following these examples: https://asciidoctor.org/docs/user-manual/#copy-and-paste-friendly-callouts

faec:
thanks for the docs! For future reference, is there a way I can preview what the generated form will look like? (I tried building the project but apparently I'm missing some dependencies.) Otherwise, is this ok to merge?

lcawl:
You can use the preview that's generated as part of the elasticsearch-ci/docs PR check. If you follow the link and look at the console output, it says:

> 13:48:19 INFO:build_docs:A preview will soon be available at http://stack-docs_453.docs-preview.app.elstc.co/diff

That URL shows you a preview of the changed file(s).

Alternatively, you can build the book on your local machine by using the commands listed here: https://github.com/elastic/docs/blob/master/doc_build_aliases.sh
In this case, since you'd be building the Stack Overview, you'd need to check out the stack-docs, elasticsearch, and kibana repos at the appropriate branch levels.

PR #467, opened 2019-08-22
[docs] Update Hadoop to use version attribute

bmorelli25:
I think the hardwiring of `7.3.0` is a bug. Everywhere else, including the header for this section, uses `{version}`.


bmorelli25:
Lisa is out. Deb or James can you take a looksie please.

PR #484, opened 2019-09-01
Azure update

alaudazzi:
Add MS Azure as new supported cloud provider.

### DIFF
--- a/docs/en/install-upgrade/installing-stack.asciidoc
+++ b/docs/en/install-upgrade/installing-stack.asciidoc
@@ -30,7 +30,7 @@ on are in place.
 
 The https://www.elastic.co/cloud/elasticsearch-service[Elasticsearch Service]
 on Elastic Cloud is the official hosted Elasticsearch and Kibana offering from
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/installing-stack.asciidoc
```diff
@@ -30,7 +30,7 @@ on are in place.
 
 The https://www.elastic.co/cloud/elasticsearch-service[Elasticsearch Service]
 on Elastic Cloud is the official hosted Elasticsearch and Kibana offering from
-Elastic. The Elasticsearch Service is available on both AWS and GCP.
+Elastic. The Elasticsearch Service is available on AWS, GCP, and Microsoft Azure.
```
nrichers: Minor: Do we want to spell out what AWS and GCP stand for since we are giving the full "Microsoft Azure" name?

alaudazzi: Good point, thanks Nik.

PR #492, opened 2019-09-05
[Docs] Added list of ML jobs and updated screenshot

benskelker:
Draft

Will add pages for each job when we get the information.
Need to improve the short description of some jobs - @randomuserid and @MikePaquette

Thanks,


spong:
Looks good @benskelker! Thanks for adding the note about how an enabled job begins processing data two weeks back as well. To clarify on the implementation, that is the max amount of time the job will go back, but if it was previously stopped less than two weeks prior, it will pick up from that point (in effort to not re-process the same data twice).

PR #498, opened 2019-09-09
[DOCS] Deletes Cannot update data frame transform limitation

szabosteve:
This PR deletes "Cannot update data frame transform" limitation because it is no longer valid.

Related issue: https://github.com/elastic/ml-team/issues/187#issuecomment-529356690

### DIFF
--- a/docs/en/stack/data-frames/limitations.asciidoc
+++ b/docs/en/stack/data-frames/limitations.asciidoc
@@ -26,12 +26,12 @@ status.
 
 [float]
 [[df-ui-limitation]]
-=== {dataframe-cap} UI will not work during a rolling upgrade from 7.2 to 7.3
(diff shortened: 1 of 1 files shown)


sophiec20:
Can I please include another update to the Limitations page..

https://www.elastic.co/guide/en/elastic-stack-overview/master/dataframe-limitations.html#df-ui-limitation

The Transforms UI will not fully work in any cluster that contains mixed node versions with 7.2. Please just explain the scenario as a `rolling upgrade from 7.2` and do not specifically name the target version (as it is any target version that supports rolling upgrades i.e. 7.3 -> 7.last)

szabosteve:
@sophiec20 fef6b71 does the change that you requested on the rolling update limitation.

szabosteve:
Backported to 7.x and 7.4.

PR #540, opened 2019-09-25
Add docs for `create_doc` index privilege

bizybot:
This commit adds documentation for new index privilege
`create_doc` which only allows indexing of new documents
but no updates to existing documents via Index or Bulk APIs.

Relates: https://github.com/elastic/elasticsearch/pull/45806


lcawl:
These source files have moved, so I recreated these changes in https://github.com/elastic/elasticsearch/pull/47584

PR #581, opened 2019-10-07
[DOCS] Preliminary work to split Infrastructure and Logs monitoring into two books

Titch990:
Partial implementation of https://github.com/elastic/observability-dev/issues/98.

For now, I've left the installation section as one topic as it doesn't seem sensible to split it then recombine it as a result of the integrations work, at least not before we have a clearer idea of the overall Observability documentation design in this area (in progress as https://github.com/elastic/observability-dev/issues/14).

Therefore, I've also left Infrastructure and Logging as one suitably named "book" for now, which we can split later.

Related issues https://github.com/elastic/observability-dev/issues/378 (the diagrams need to be updated). Since that issue is slightly wider, and covers a bit more than just splitting Metrics and Logs, I've just duplicated the existing diagrams for now in the two split topics.


Titch990:
Hi @elastic/infrastructure. Could I get a review on this, please? Docs build can be seen here http://stack-docs_581.docs-preview.app.elstc.co/guide/en/infrastructure/guide/master/index.html

andresrc:
@bmorelli25 My understanding is that this PR explicitly says that is not closing the issue, but just providing the starting point of it.

Anyway, the comment raises another concern: if the intermediate state is one that we don't want published, we may either continue the work in the same PR or group the changes in a feature branch.

bmorelli25:
Ultimately, @Titch990 owns this content. If y'all decide to merge this as-is, that's totally cool with me.

If we decide not to merge this yet, I don't think we need a working branch. This PR is only a few lines away from being "done".

My main concern is that the Kibana docs redesign is on track for 7.5. So by then, we need somewhere to put all of the Observability UI documentation.

> For now, I've left the installation section as one topic as it doesn't seem sensible to split it then recombine it as a result of the integrations work, at least not before we have a clearer idea of the overall Observability documentation design in this area (in progress as elastic/observability-dev#14).

If we're not going to follow the original plan, that's fine, we just need to make sure to involve `\@gchaps`  and `\@tbragin`  in those conversations... and the conversations need to start ASAP.

Titch990:
@bmorelli25 Sorry, perhaps I wasn't clear. 

I would really have preferred to address this after: 

(a) We have solved the TOC navigation issue.
(b) We have a better idea of how the integrations are going to fit into the Observability documentation. 

But . . . 

I'd started this work before I realised the importance of these other two things and how essential they were for the final implementation. So I needed to park it neatly somewhere, and I realised I could complete a partial implementation.

I can hold off merging this if you think it makes things worse. But I thought it made things a bit better, and closer to where we need to be. We now have separate introductory topics for Metrics and for Logs, whereas before they were mixed in one topic, which sometimes referred to only Metrics and sometimes to both Logs and Metrics. There are some changes under the hood, but that's actually the main change here. Everything else is as-is.

Integrations need to be referenced by both Logs and Metrics, so don't belong in either the Logs or Metrics "books". And the first ones are coming soon, so need a place to go. With the navigation situation as it is at present, I didn't want to add yet another "book" for Integrations, then have to provide confusing cross-book links that will definitely need to be followed out and back in a basic user journey.

My (very) short term plan is to keep the one book for Logs and Metrics, and add an Integrations section in this "book". Then I can split the "Setting up" topic (that's hard to split at present) into topics for Metrics and for Logs, each referencing the appropriate content in the Integrations section. Then, I can make separate sub-sections for the Metrics and Logs content. I really want to hold out splitting this into separate books until the navigation issue is resolved, but it will be ready to split as soon as it is.

I wasn't aware of the timescale for the Kibana docs redesign, nor that it would mean the existing logs and metrics sections would need to be moved elsewhere immediately. If that is the case, I don't think it's a big deal, because I can simply swap the placeholder topics in this book with the current set of Kibana topics. So instead of stub topics in this book that reference the fuller descriptions in the Kibana book, the stub topics will be in Kibana and the fuller descriptions will be here.

So now all I need to do is chase up the timescale for the improved navigation solution. As far as I can see, there has been no action on that issue yet.

Titch990:
So, @bmorelli25 @andresrc. What do you think?

My gut feeling is to publish this, add an Integrations section as required by https://github.com/elastic/observability-dev/issues/307, then split the rest into separate Metrics and Logging sections. With that in place it will be easier to split the "Set up" section in a way that's intuitive for the user, and will work with the integrations as they are produced. The Kibana orphan topics will also have an obvious home.

Publishing this as is, while not ideal, unblocks several other tasks, and reduces the dependence on a speedy solution for the navigation issues (although I shall still keep plugging away at that, because it's still a problem for the rest of Observability too).

bmorelli25:
> My (very) short term plan is to keep the one book for Logs and Metrics, and add an Integrations section in this "book". Then I can split the "Setting up" topic (that's hard to split at present) into topics for Metrics and for Logs, each referencing the appropriate content in the Integrations section. Then, I can make separate sub-sections for the Metrics and Logs content. I really want to hold out splitting this into separate books until the navigation issue is resolved, but it will be ready to split as soon as it is.

I can see why you'd want to go this route, but it isn't the plan that was decided upon in elastic/observability-dev#14. I'm not married to the plan we came up with, but if you want to make changes like this, we need to discuss them with the right people first. My fear with this plan is it will further entangle these docs. Right now, the books are small -- it's an easy separation to make. The longer the content is able to grow together, the more entangled it becomes. Adding sub-sections and parts for Metrics/Logs will further that problem. In addition, we end up with a weird mix of some Observability docs together and others separate.

> I wasn't aware of the timescale for the Kibana docs redesign, nor that it would mean the existing logs and metrics sections would need to be moved elsewhere immediately. If that is the case, I don't think it's a big deal, because I can simply swap the placeholder topics in this book with the current set of Kibana topics. So instead of stub topics in this book that reference the fuller descriptions in the Kibana book, the stub topics will be in Kibana and the fuller descriptions will be here.

I think the problem with this is you're trying to solve our navigation issue with the structure of the docs. That isn't the right approach for our doc architecture. I know our nav isn't perfect right now, but it's something we have to live with.

dedemorton:
@brandon asked me to weigh in on this discussion because I've been involved in logs and metrics since the beginning...even before the apps existed.

TBH, I'm on the fence as to whether this content belongs in a single book or two (mainly because the book paradigm is wrong for online content). If you're going to put everything into one book, though, a flat structure with the current level of repetition is problematic. The relationship between topics needs to be obvious when users look at the TOC. Flipping through topics one after another that contain a lot of duplicated content, but vary slightly, is not a great user experience, IMO, especially if it's expected that some users won't be running both UIs. 

What I'd really like to see is a structure that's organized around user goals with content that describes how users achieve those goals. It would also be good to see how the content in the Kibana docs (which I assume you're moving) fits into the content that you have here. I suspect the flat structure will not scale. I learned the hard way with Beats that having a structure that scales (even when you have a small amount of content) is pretty important.

I'd suggest developing an outline that includes the content we have now plus all the content that needs to be written. This might help you avoid some rework in the future. It is a major pain to reorganize asciidoc content because the hierarchy is not decoupled from the content. If you can get the structure mostly right in the beginning, it will save you a lot of rework later on.

Titch990:
@dedemorton 

> the book paradigm is wrong for online content

I couldn't agree more!

> I'd suggest developing an outline that includes the content we have now plus all the content that needs to be written. 

This is exactly what I am trying to do. However, it appears that this split needs to be done first. 

> It is a major pain to reorganize asciidoc content because the hierarchy is not decoupled from the content.

Not directly related to this pull request, I know, but I wonder if there is something we can do to help ourselves? Firstly, let's assume that each asciidoc file corresponds directly to one online topic. I know that's not always the case, but it is usually, so let's pretend.

Within a topic, the reader doesn't care what the the heading level numbers are for the topic heading or the section headings. They simply care that there are headings. The topic heading helps them decide if the topic is relevant to them. The section headings divide up the content and make it easier to read or to scan.

The topic heading needs to be a proper asciidoc coded heading, at the appropriate level, because as you say, the heading level determines where in the hierarchy the topic is located. But what if we could define some other style for the section headings (even a simple *bold* would be better than nothing)? The section headings just need to stand out from the body text. They don't need to be an actual heading style. Then, if the content hierarchy changes, the only thing that needs to change is the level of the topic heading. The section headings can stay exactly as they are, because they are not heading levels.

Titch990:
OK, we've decided to go for the full split, so this is now part of a multi-repo change (see elastic/observability-dev#98). DO NOT PUSH UNTIL EVERYTHING IS READY.

Titch990:
Note that this is one of four linked PRs, which need to be merged together to avoid build failures.

elastic/stack-docs#581 (this one)
elastic/docs#1312
elastic/kibana#48633
elastic/beats#14158

DO NOT MERGE UNTIL EVERYTHING IS READY

dedemorton:
Can you please assign @zuketo as the reviewer here since it pertains to info architecture changes?

Titch990:
Can I have a review please?

Note that this is one of four linked PRs, and this does not build on its own. Because it doesn't build in isolation, I can't provide a link to the output for review, and because of the changes and renames, it's hard/impossible to review properly from the sources. However, I think there is value in a proper review of some of the changed topics.

So I've done screenshots of the output from my local "everything together" build. Here's what there is and here's what I think is helpful to review (some topics are unchanged except for the location).

The screenshots are here: https://drive.google.com/open?id=18hyRbxgh80ITyLXoQT4Z82P7k2zmaa6n 

- 01-top-level.png - showing what's changed in the top level navigation
- 02-logs-mon-guide.png - the new Logs "book"
- 03-logs-mon-guide-overview.png - the Logs overview (split from a single topic that gave an overview of both metrics and logs)
- 04-set-up-logs-mon.png - setting up Logs (split from a single topic that gave set-up instructions for both metrics and logs)
- 05-logs-app.png - a brief description of the Logs app. Unchanged
- 06-metrics-mon-guide.png - the new Metrics "book"
- 07-metrics-mon-overview.png -the Metrics overview (split from a single topic that gave an overview of both metrics and logs)
- 08-set-up-metrics-mon.png Setting up Metrics (split from a single topic that gave set-up instructions for both metrics and logs)
- 09-metrics.png - unchanged. A list of some metrics values and what they mean
- 10-metrics-app.png- a brief description of the Metrics app. Unchanged

Outstanding changes. 

- The architecture diagrams in the two overview topics still reference both Logs and Metrics in each case. There is a separate ticket to fix this https://github.com/elastic/observability-dev/issues/378
- There are still some instances of "Infrastructure" that need to be replaced by "Metrics". There is also a ticket for this https://github.com/elastic/observability-dev/issues/355

Titch990:
@zuketo feel free to review and comment if you want to. However, this work is holding up other docs work for the 7.5 release. So if we are going to revisit the decision to go ahead with the split, that decision needs to be made promptly, so I can pull this and tackle other outstanding issues in a different way.

docs/en/logs/index.asciidoc
```diff
@@ -0,0 +1,17 @@
+:doctype: book
+:metrics: infrastructure
+:metrics-app: Infrastructure app
+:logs: logs
+:logs-app: Logs app
```
zuketo: Should App be capitalized? If so, this is probably a separate issue as it applies to more than just the Logs App (and this PR). https://wiki.elastic.co/display/MAR/Term+usage+guide

docs/en/logs/logs-installation.asciidoc
```diff
@@ -0,0 +1,92 @@
+[[install-logs-monitoring]]
+[role="xpack"]
+== Setting up logs monitoring
+
+To set up logs monitoring, you need:
+
+* An Elasticsearch cluster and Kibana (version 6.5 or later) with a basic license
```
zuketo: Can we leave out the license part, or describe it as the default distribution?

An Elasticsearch cluster and Kibana (version 6.5 or later).

or 

An Elasticsearch cluster and Kibana (ensure it is Elastic's default distribution, version 6.5 or later).

Titch990: This is a wider issue than this PR. This PR simply relates to the work to split the existing Infrastructure Monitoring Guide "book" into two separate "books" for Metrics and Logs, so that text was already there as written.

Firstly, I'm not sure that this statement is consistently applied in other places where there is a dependency on the license type and/or the version.

Secondly, is it even appropriate to be mentioning older versions of the Elastic stack in the 7.5 documentation? It's quite possible that some of the functionality described for the Logs app in the 7.5 version of the docs won't be available in earlier versions, but by stating an earlier version as a prerequisite, we imply that it will work as described. 

It may be appropriate to mention a version when a feature is first introduced (I think Logs was introduced in 6.5?) but I'm not sure it adds value after that?

docs/en/logs/index.asciidoc
Titch990: Thanks for doing the review at short notice! 

No, I don't think "app" should be capitalised. In general, we avoid capitals unless things really are proper names, or there is anther very good reason for using them. As far as I'm aware, using "app" agrees with the marketing guidelines.

docs/en/logs/logs-installation.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+// Add a link to what constitutes a basic license. And is this any different for the cloud?
+
+* Appropriate {beats} shippers (version 6.5 or later) installed and enabled on each system you want to
```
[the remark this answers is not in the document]
Titch990: This content predates the split which is the focus of this PR. Raised https://github.com/elastic/stack-docs/issues/646 to address it.

docs/en/logs/logs-installation.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+[float]
+==== Install {beats} yourself
+
+If your data source doesn't have a {beats} module, or if you want to install {beats} the old fashioned way, follow the instructions in {filebeat-ref}/filebeat-modules-quickstart.html[{filebeat} modules quick start] and enable modules for the logs you want to collect.
+If there is no module for the logs you want to collect, see the {filebeat-ref}/filebeat-getting-started.html[{filebeat} getting started] to learn how to configure inputs.
+
+[float]
+=== Enable modules
+
+However you install {beats}, you need to enable the appropriate modules in {filebeat} to start collecting logs data.
```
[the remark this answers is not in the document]
Titch990: This content predates the split which is the focus of this PR. Raised https://github.com/elastic/stack-docs/issues/646 to address it.

docs/en/logs/logs-installation.asciidoc
zuketo: Yup, agreed, this is a larger issue that should be independent of this PR.

Titch990:
Jenkins retest please

docs/en/logs/logs-installation.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+
+// ++ I think some of this is still necessary even if you've followed the instructions to install Beats from Kibana.
+// ++ The instructions there explain how to enable the module. Below, we enable more stuff.
+// ++ What about if you are using Cloud? Is anything different?
```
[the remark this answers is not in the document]
Titch990: Thanks for the info.

Titch990:
@weltenwort I thought I got all the instances of non-standard. I found another and nuked it!

I think for now I'll stick with "Logs monitoring" throughout. The name of the app is the "Logs app" according to the marketing guide, and we also have the "Metrics app" and "Metrics monitoring guide". I'm happy to revisit if anyone feels strongly, but at least for now, we are consistent.

nik9000:
I've talked with @Titch990 and we understand the CI failure that is on this and will handle it after merging. This is one of those unfortunate times when CI is causing us a fair bit of trouble because we can't link changes across repos.

Titch990:
I know this doesn't build. It's part of a bigger inter-repo change and it won;t build until the other changes are made too.

Issue #455, opened 2019-08-06
Create a "one-pager" for running Elasticsearch and Kibana in Docker w Security Enabled


debadair:
"Specifically looking for getting started instruction smoothness for Docker." (@derickson)

debadair:
After discussing with @jethr0null, getting some more background from @derickson, and looking at our existing docs, Docker Hub image readmes, Open Distro docs, third party tutorials, and forum threads, there are a few key issues:

- Nowhere do we officially document how to run ES and Kibana in Docker containers and get them talking to one another. (https://discuss.elastic.co/t/kibana-docker-image-doesnt-connect-to-elasticsearch-image/79511)
- The instructions for running with Docker aren't consistent. For example, we talk about running ES with Docker Compose, but we don't provide a sample compose file for Kibana. 
- Our sample compose file starts two ES instances. That seems like a pattern that we probably don't want to encourage, especially in a section titled "Production Mode".
- You have to click around to get to the info about enabling security. 

We might be able to repurpose some of the instructions @jethr0null wrote for the Docker Hub image readmes. The Discuss thread & referenced issue have info about getting Kibana and ES talking to one another. 

I'm also going to follow up with @woodywalton, who has been looking at collecting this information in the context that prompted the ask.

PR #586, opened 2019-10-08
[DOCS] Added instructions for 3 nodes + kibana

debadair:
Uses a compose file that launches 3 nodes + Kibana. 

Related to #455

### DIFF
--- a/docs/en/getting-started/docker-compose.yml
+++ b/docs/en/getting-started/docker-compose.yml
@@ -5,9 +5,9 @@ services:
     container_name: es01
     environment:
       - node.name=es01
+      - cluster.name=es-docker-cluster
       - discovery.seed_hosts=es02,es03
       - cluster.initial_master_nodes=es01,es02,es03
-      - cluster.name=docker-cluster
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
     ulimits:
@@ -15,19 +15,20 @@ services:
         soft: -1
         hard: -1
     volumes:
-      - esdata01:/usr/share/elasticsearch/data
+      - data01:/usr/share/elasticsearch/data
     ports:
       - 9200:9200
     networks:
-      - esnet
+      - esnet    
+
   es02:
     image: docker.elastic.co/elasticsearch/elasticsearch:7.4.0
     container_name: es02
     environment:
       - node.name=es02
+      - cluster.name=es-docker-cluster
       - discovery.seed_hosts=es01,es03
       - cluster.initial_master_nodes=es01,es02,es03
-      - cluster.name=docker-cluster
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
     ulimits:
@@ -35,19 +36,20 @@ services:
         soft: -1
         hard: -1
     volumes:
-      - esdata02:/usr/share/elasticsearch/data
+      - data02:/usr/share/elasticsearch/data
     ports:
       - 9201:9201
     networks:
       - esnet
+  
   es03:
     image: docker.elastic.co/elasticsearch/elasticsearch:7.4.0
     container_name: es03
     environment:
       - node.name=es03
+      - cluster.name=es-docker-cluster
       - discovery.seed_hosts=es01,es02
       - cluster.initial_master_nodes=es01,es02,es03
-      - cluster.name=docker-cluster
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
     ulimits:
@@ -55,18 +57,17 @@ services:
         soft: -1
         hard: -1
     volumes:
-      - esdata03:/usr/share/elasticsearch/data
+      - data03:/usr/share/elasticsearch/data
     ports:
       - 9202:9202     
     networks:
       - esnet
+
   kib01:
     image: docker.elastic.co/kibana/kibana:7.4.0
     container_name: kib01
     ports:
       - 5601:5601
-    expose:
-      - "5601"
     environment:
       ELASTICSEARCH_URL: http://es01:9200
       ELASTICSEARCH_HOSTS: http://es01:9200
@@ -74,11 +75,11 @@ services:
       - esnet      
 
 volumes:
-  esdata01:
+  data01:
     driver: local
-  esdata02:
+  data02:
     driver: local
-  esdata03:
+  data03:
     driver: local    
 
 networks:
--- a/docs/en/getting-started/docker-compose.yml
+++ b/docs/en/getting-started/docker-compose.yml
@@ -1,0 +1,85 @@
+version: '2.2'
+services:
+  es01:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.4.0
+    container_name: es01
+    environment:
+      - node.name=es01
+      - discovery.seed_hosts=es02,es03
+      - cluster.initial_master_nodes=es01,es02,es03
+      - cluster.name=docker-cluster
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - esdata01:/usr/share/elasticsearch/data
+    ports:
+      - 9200:9200
+    networks:
+      - esnet
+  es02:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.4.0
+    container_name: es02
+    environment:
+      - node.name=es02
+      - discovery.seed_hosts=es01,es03
+      - cluster.initial_master_nodes=es01,es02,es03
+      - cluster.name=docker-cluster
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - esdata02:/usr/share/elasticsearch/data
+    ports:
+      - 9201:9201
+    networks:
+      - esnet
+  es03:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.4.0
+    container_name: es03
+    environment:
+      - node.name=es03
+      - discovery.seed_hosts=es01,es02
+      - cluster.initial_master_nodes=es01,es02,es03
+      - cluster.name=docker-cluster
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - esdata03:/usr/share/elasticsearch/data
+    ports:
+      - 9202:9202     
+    networks:
+      - esnet
+  kib01:
+    image: docker.elastic.co/kibana/kibana:7.4.0
+    container_name: kib01
+    ports:
+      - 5601:5601
+    expose:
+      - "5601"
+    environment:
+      ELASTICSEARCH_URL: http://es01:9200
+      ELASTICSEARCH_HOSTS: http://es01:9200
+    networks:
+      - esnet      
+
+volumes:
+  esdata01:
+    driver: local
+  esdata02:
+    driver: local
+  esdata03:
+    driver: local    
+
+networks:
+  esnet:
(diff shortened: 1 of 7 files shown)


docs/en/getting-started/docker-compose.yml
```diff
@@ -0,0 +1,85 @@
...
+      - 9202:9202     
+    networks:
+      - esnet
+  kib01:
+    image: docker.elastic.co/kibana/kibana:7.4.0
+    container_name: kib01
+    ports:
+      - 5601:5601
+    expose:
+      - "5601"
```
dliappis: Is `expose` serving a specific need? My understanding from [the docs](https://docs.docker.com/compose/compose-file/compose-file-v2/#expose) is that it is needed by the [legacy container links](https://docs.docker.com/network/links/) but as we use a user-defined bridge network this shouldn't be needed.

docs/en/getting-started/docker-compose.yml
```diff
@@ -0,0 +1,85 @@
...
+volumes:
+  esdata01:
+    driver: local
+  esdata02:
+    driver: local
+  esdata03:
+    driver: local    
+
+networks:
+  esnet:
```
dliappis: nit: add a newline here

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -2,9 +2,34 @@
...
+["source","yaml",subs="attributes"]
+--------------------------------------------
+include::images/docker-compose.yml[]
+--------------------------------------------
+
+. Run `docker-compose` to bring up a three node Elasticsearch cluster and Kibana:
++
+["source","sh",subs="attributes"]
+--------------------------------------------
+docker-compose up
```
dliappis: Does it make sense to document the link (http://localhost:5601) to visit Kibana and continue from there?

jethr0null:
@fxdgear if you have a moment, would you mind giving this a look over as well?

jethr0null:
@debadair I haven't had a chance to test/thoroughly review yet but one thing I notice is that we can eliminate the explicit docker pull steps if we want given that the images are specified in the docker-compose files (and the version gets supplied when the command is run). I don't personally see any benefit in explicitly calling out the pull step and it would allow us to reduce the number of documented steps/user actions if we remove.

fxdgear:
From just a quick glance nothing seems out of place. 

I would ultimately recommend using "env" files for the envs for both the ssl'd versions and the non ssl'd versions this would allow for some de-duplication of work.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
 [[get-started-docker]]
```
jethr0null: I wasn't able to run the non-secured example with the default docker for mac settings (2 CPUs and 2 GiB Memory) without at least one of the nodes failing. I bumped up the allocation to 4 CPU and 4 GiB Memory (Docker for Mac -> Settings -> Advanced) and tested and it appears to work fine. I'm thinking we might want to add a prerequisite section that outlines the resource requirements for running these configurations (while my test worked, it would be good to have someone from the ES team weigh in on optimal recommendations).

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+=== Run with Docker Compose
+
+To get the default distributions of {es} and {kib} up and running in Docker, 
+you can use `docker-compose`:
+
+. Pull the docker images for {es} and {kib}:
++
+["source","sh",subs="attributes"]
+--------------------------------------------
+docker pull docker.elastic.co/elasticsearch/elasticsearch:{version}
```
jethr0null: We can eliminate the explicit docker pull steps if we want given that the images are specified in the docker-compose files (and the version gets supplied when the command is run). I don't personally see any benefit in explicitly calling out the pull step and it would allow us to reduce the number of documented steps/user actions if we remove.

docs/en/getting-started/get-started-docker.asciidoc
dliappis: I agree we should clarify that the user needs to check that there's sufficient memory in Docker for Mac (I suppose also for Windows) and we recommend at least 4GiB -- given that these examples have 512MiB heap (plus half of that for Max Direct Memory) x 3 + enough memory for Kibana. Also we should include the Docker link that it can be changed via https://docs.docker.com/docker-for-mac/#resources

debadair: In my own testing, once I bumped to 4 GiB things worked fine. Agree that needs to be called out.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+--------------------------------------------
+
+. Create a `docker-compose.yml` file for the Elastic Stack:
++
+["source","yaml"]
+--------------------------------------------
+include::docker/docker-compose.yml[]
+--------------------------------------------
+
+. Run `docker-compose` to bring up a three node Elasticsearch cluster and Kibana:
```
jrodewig:
```suggestion
. Run `docker-compose` to bring up a three node {es} cluster and {kib}:
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+--------------------------------------------
+
+. Submit a `_cat/nodes` request to see that the nodes are up and running:
++
+[source,console]
+--------------------------------------------------
+GET /_cat/nodes?v
+--------------------------------------------------
+
+. Open Kibana to load sample data and interact with the cluster:
```
jrodewig:
```suggestion
. Open {kib} to load sample data and interact with the cluster: 
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+
+. Bring up the three node {es} cluster:
++
+--
+["source","sh"]
+----
+docker-compose -f elastic-docker-tls.yml up -d
+----
+
+IMPORTANT: At this point, Kibana cannot connect to the {es} cluster.
```
jrodewig:
```suggestion
IMPORTANT: At this point, {kib} cannot connect to the {es} cluster.
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+You must generate a password for the built-in `kibana` user, update the `ELASTICSEARCH_PASSWORD` 
+in the compose file, and restart to enable Kibana to communicate with the secured cluster.
```
jrodewig:
```suggestion
in the compose file, and restart to enable {kib} to communicate with the secured cluster. 
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
++
+--
+["source","txt",subs="-callouts"]
+----
+include::docker/elastic-docker-tls.yml[]
+----
+
+--
+
+. Use `docker-compose` to restart the cluster and Kibana:
```
jrodewig:
```suggestion
. Use `docker-compose` to restart the cluster and {kib}:
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
++
+--
+["source","sh"]
+----
+docker-compose down
+docker-compose -f elastic-docker-tls.yml up -d
+----
+--
+
+. Open Kibana to load sample data and interact with the cluster:
```
jrodewig:
```suggestion
. Open {kib} to load sample data and interact with the cluster: 
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+https://localhost:5601. 
++
+NOTE: Because SSL is also enabled for communications between {kib} and client browsers, 
+you must access {kib} via the HTTPS protocol. 
+
+[float]
+[[load-settings-file]]
+==== Loading settings from a file
+
+Specifying settings for {es} and {kibana} directly in the compose file is a convenient way to get started,
```
jrodewig: This one won't render correctly as-is.
```suggestion
Specifying settings for {es} and {kib} directly in the compose file is a convenient way to get started, 
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
-{stack}. For more information, see:
+The Elastic Docker registry contains Docker images for all the products in 
+the {stack}: https://www.docker.elastic.co/
+
+[float]
+[[run-stack-docker]]
+=== Run with Docker Compose
+
+To get the default distributions of {es} and {kib} up and running in Docker, 
+you can use `docker-compose`:
```
jrodewig: > `docker-compose`

Not sure if the literals are needed here. I'd use "Docker Compose" but don't feel strongly.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+
+. Pull the docker images for {es} and {kib}:
```
jrodewig: > docker images 

Should Docker be capitalized here? "container image" would work fine too.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+--------------------------------------------
+
+. Run `docker-compose` to bring up a three node Elasticsearch cluster and Kibana:
++
+["source","sh",subs="attributes"]
+--------------------------------------------
+VERSION={version} docker-compose up
+--------------------------------------------
+
+. Submit a `_cat/nodes` request to see that the nodes are up and running:
```
jrodewig: We may want to link to the cat nodes API here.

It also may be confusing to include a console snippet *before* we explicitly tell people to open up Kibana.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+--------------------------------------------------
+
+. Open Kibana to load sample data and interact with the cluster: 
+http://localhost:5601. 
+
+[float]
+[[get-started-docker-tls]]
+=== Run in Docker with TLS enabled 
+
+When security is enabled with a Gold or Platinum license,
```
jrodewig: > Gold or Platinum license
links?

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
+
+. Open Kibana to load sample data and interact with the cluster: 
+http://localhost:5601. 
+
+[float]
+[[get-started-docker-tls]]
+=== Run in Docker with TLS enabled 
+
+When security is enabled with a Gold or Platinum license, 
+TLS encryption must be configured for the {es} transport layer.
```
jrodewig: > TLS encryption
It may be good to spell out "Transport Layer Security (TLS)"

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,10 +1,220 @@
...
++
+--
+["source","sh"]
+----
+docker-compose -f create-certs.yml run --rm create_certs
+----
+
+--
+
+. Bring up the three node {es} cluster:
```
jrodewig:
```suggestion
. Bring up the three-node {es} cluster:
```

docs/en/getting-started/get-started-docker.asciidoc
debadair: In the getting started content, I'm opting to minimize linking folks out of the flow.

docs/en/getting-started/get-started-docker.asciidoc
jrodewig: Can discard.

docs/en/getting-started/get-started-docker.asciidoc
jrodewig: Makes sense. Thanks!

docs/en/getting-started/get-started-docker.asciidoc
debadair: Here I agree that we want to xref the subscriptions page, as they might have questions about the license levels before they dive into the procedure.

debadair:
> From just a quick glance nothing seems out of place.
> 
> I would ultimately recommend using "env" files for the envs for both the ssl'd versions and the non ssl'd versions this would allow for some de-duplication of work.

While I agree this makes sense from a maintenance standpoint, my inclination is to keep the "basic" version as simple as possible.

PR #619, opened 2019-10-14
[DOCS] Marks DFA and the related analytics types as experimental feataures

szabosteve:
This PR marks the data frame analytics and the related analytics types as experimental features.


szabosteve:
Backported to 7.x and 7.4.

PR #637, opened 2019-10-17
Use new repo path attributes

nik9000:
Mostly uses `{repname-root}` instead of `../../reponame` to find the
root of repositories. Because `../../` is confusing.

### DIFF
--- a/docs/en/getting-started/index.asciidoc
+++ b/docs/en/getting-started/index.asciidoc
@@ -11,11 +11,8 @@
 :blog-ref:          https://www.elastic.co/blog/
 :wikipedia:         https://en.wikipedia.org/wiki
 
-:kib-repo-dir:      {docdir}/../../../../kibana/docs
-:xes-repo-dir:      {docdir}/../../../../elasticsearch/x-pack/docs/en
-
-include::{asciidoc-dir}/../../shared/versions/stack/{source_branch}.asciidoc[]
-include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
+include::{docs-root}/shared/versions/stack/{source_branch}.asciidoc[]
+include::{docs-root}/shared/attributes.asciidoc[]
 
 include::get-started-stack.asciidoc[]
 include::get-started-docker.asciidoc[]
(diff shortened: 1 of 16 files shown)


nik9000:
Another one for you @lcawl ! This one's diff should be empty.

docs/en/getting-started/index.asciidoc
```diff
@@ -11,11 +11,8 @@
 :blog-ref:          https://www.elastic.co/blog/
 :wikipedia:         https://en.wikipedia.org/wiki
 
-:kib-repo-dir:      {docdir}/../../../../kibana/docs
```
lcawl: Were these attributes intentionally removed?  I would expect them to just be updated with {stack-docs-root} paths

docs/en/glossary/index.asciidoc
```diff
@@ -6,10 +6,10 @@
 :logstash-terms:       true
 :xpack-terms:          true
 
-include::{asciidoc-dir}/../../shared/versions/stack/{source_branch}.asciidoc[]
-include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
+include::{docs-root}/shared/versions/stack/{source_branch}.asciidoc[]
+include::{docs-root}/shared/attributes.asciidoc[]
 
-:es-repo-dir:          {docdir}/../../../../elasticsearch/docs/reference
+:es-glossary:          {elasticsearch-reference}/glossary.asciidoc
```
lcawl: I think in this context it makes me rethink the name of this attribute. Is it too late to rename it to something like "es-ref-root" or "es-ref-dir" or "es-reference-dir"? Then it's clearer that it's referring to a path, not to the book or its URL
```suggestion
:es-glossary:          {elasticsearch-reference}/glossary.asciidoc
```

nik9000: https://github.com/elastic/docs/pull/1316

docs/en/getting-started/index.asciidoc
nik9000: I switched everything to `{kibana-root}/docs`. I didn't think we needed the extra indirection there. Is that ok?

docs/en/glossary/index.asciidoc
lcawl: Thanks!

nik9000:
Woah! I think this one isn't going to happen. It's been years!

PR #641, opened 2019-10-22
[DOCS] Specifies that regression is a batch analysis

szabosteve:
This PR makes it clear that regression is a batch analysis.

### DIFF
--- a/docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
+++ b/docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
@@ -75,6 +75,9 @@ predictions are combined. The ensemble learning technique that we use in the
 {stack} is a type of boosting called extreme gradient boost (XGboost) which 
 combines decision trees with gradient boosting methodologies.
 
+{regression-cap} works as a batch analysis, if new data comes into your index, 
+you have to perform the {reganalysis} again on the altered data.
+
  
 [discrete]
 [[dfa-regression-evaluation]]


docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
```diff
@@ -75,6 +75,9 @@ predictions are combined. The ensemble learning technique that we use in the
 {stack} is a type of boosting called extreme gradient boost (XGboost) which 
 combines decision trees with gradient boosting methodologies.
 
+{regression-cap} works as a batch analysis, if new data comes into your index,
```
lcawl:
```suggestion
{regression-cap} works as a batch analysis. If new data comes into your index, 
```

docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
```diff
@@ -75,6 +75,9 @@ predictions are combined. The ensemble learning technique that we use in the
 {stack} is a type of boosting called extreme gradient boost (XGboost) which 
 combines decision trees with gradient boosting methodologies.
 
+{regression-cap} works as a batch analysis, if new data comes into your index, 
+you have to perform the {reganalysis} again on the altered data.
```
lcawl: > ...on the altered data.

This seems to imply that you can/should re-run the analysis only on the data that's changed. It doesn't specify how you'd make the job aware of which subset to re-analyze. Can we clarify?

If, on the other hand, you have to run re-run the job against the full set of data, this sentence should change to something like this: "If new data comes into your index, you must restart the {dfanalytics-job}".

szabosteve:
Backported to 7.x, 7.5, and 7.4.

PR #654, opened 2019-10-28
[DOCS] [FOR 7.5] Add superseded banner to older versions of Infrastructure Monitoring Guide

Titch990:
DO NOT MERGE UNTIL 7.4 IS NO LONGER THE CURRENT RELEASE!

At the same time, also needs to be back ported to all earlier branches in which this doc exists (7.3 down to 6.5)

### DIFF
--- a/docs/en/infraops/page_header.html
+++ b/docs/en/infraops/page_header.html
@@ -1,5 +1,5 @@
 You are looking at documentation for an older release.
-Not what you want?
-In the current release, the Infrastructure monitoring guide has been replaced by two documents, the
+Starting in version 7.5, see the
 <a href="https://www.elastic.co/guide/en/logs/guide/current/index.html"> Logs monitoring guide</a>
-and the <a href="https://www.elastic.co/guide/en/metrics/guide/current/index.html"> Metrics monitoring guide</a>.
+and the <a href="https://www.elastic.co/guide/en/metrics/guide/current/index.html"> Metrics monitoring guide</a>
+for information about infrastructure monitoring.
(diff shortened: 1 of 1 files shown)


Titch990:
@elasticmachine run elasticsearch-ci/docs rebuild

docs/en/infraops/page_header.html
```diff
@@ -0,0 +1,5 @@
+You are looking at documentation for an older release.
+Not what you want?
+In the current release, the Infrastructure monitoring guide has been replaced by two documents, the
+<a href="https://www.elastic.co/guide/en/logs/guide/current/index.html"> Logs monitoring guide</a>
+and the <a href="https://www.elastic.co/guide/en/metrics/guide/current/index.html"> Metrics monitoring guide</a>.
```
debadair: I'd avoid referencing current and be more explicit. Something like:

> Starting in version 7.5, see the Logs monitoring guide and the Metrics monitoring guide for information about infrastructure monitoring.

With the appropriate links, of course. 



>

Titch990: @debadair Agreed, except I think the links probably still need to be to the current versions? If the reader finds the wrong docs, when we direct them to the right docs, the most helpful version to take them to initially is probably the current version? From there they can navigate to another version if they want to.

I'm just checking this because you seem to be saying that I need to change the links here too. At present, they do go to the current version. Am I missing something?

bmorelli25:
Rerunning the doc build since it hasn't run in a while.

@elasticmachine, run elasticsearch-ci/docs

PR #661, opened 2019-10-29
[DOCS] Renaming references to Infrastructure app to Metrics where appropriate

Titch990:


docs/en/logs/logs-overview.asciidoc
```diff
@@ -4,9 +4,8 @@
 
 Logs monitoring enables you to view logs from your infrastructure and identify problems in near real time.
 You can view logs from servers, containers, services, and so on.
-Then you can drill down to view more detailed information about an individual log entry, or you can seamlessly switch to view corresponding metrics, uptime information or APM traces where available.
+Then you can drill down to view more detailed information about an individual log entry, or you can seamlessly switch to view corresponding metrics, uptime information or APM traces where available. You can also use machine learning to automatically detect some kinds of log anomalies.
```
sgrodzicki: Machine learning is only available in Platinum. Do we still want to highlight it here?

sorantis: That depends on how we treat other Platinum content. [The ML docs](https://www.elastic.co/guide/en/elastic-stack-overview/current/ml-dfanalytics.html) are tagged with `x-pack`, but so is our [Logs UI](https://www.elastic.co/guide/en/logs/guide/current/logs-app-overview.html).

Titch990: I've just been working on the Analysis tab, and we've agreed to state there that the Analysis tab is only available if "the machine learning feature is enabled". Apparently we avoid mentioning licensing requirements, as they can change, so it's preferable to mention dependencies. I'll add that here too.

PR #690, opened 2019-11-19
[DOCS] Adds classification example to the DFA documentation

szabosteve:
This PR adds a classification worked example to the data frame analytics documentation that uses the Kibana sample data flight dataset.

Related issue: https://github.com/elastic/machine-learning-data/issues/275
Related content: https://gist.github.com/Winterflower/e2f226fd615a556bc1ddf9745d68df15


docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+
+[source,console]
+--------------------------------------------------
+POST _ml/data_frame/analytics/model-delayed-flights/_start
+--------------------------------------------------
+// TEST[skip:TBD]
+
+
+The job takes a few minutes to run. Runtime depends on the local hardware and 
+also on the number of documents and fields that analyzed. The more fields and
```
przemekwitek: `... documents and fields that analyzed`

How about changing to "that are analyzed" or similar?

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+[source,console]
+--------------------------------------------------
+POST _ml/data_frame/analytics/model-delayed-flights/_start
+--------------------------------------------------
+// TEST[skip:TBD]
+
+
+The job takes a few minutes to run. Runtime depends on the local hardware and 
+also on the number of documents and fields that analyzed. The more fields and 
+documents, the longer the job to run.
```
przemekwitek: `... the longer the job to run`

Please check if this is correct English.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+----
+<1> The `dependent_variable` with the ground truth value. This is what we are 
+trying to predict with the {classanalysis}.
+<2> The top classes object contains the predicted classes with the highest 
+probability.
+<3> The probability that the datapoint belongs to the given class. This is a 
+value between 0 and 1. The higher the number the higher the probability that the 
+datapoint belongs to the class.
+<4> Name of the class.
+<5> The prediction. The field name is suffixed with `_prediction`.
```
przemekwitek: This is true unless the user specified `prediction_field_name` via the API. In such a case, there will be no `_prediction` suffix.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+              {
+                "class_probability" : 0.06066463494150393,
+                "class_name" : "true"
+              }
+            ],
+            "FlightDelay_prediction" : "false", <5>
+            "is_training" : false <6>
+          }
+----
+<1> The `dependent_variable` with the ground truth value. This is what we are
```
przemekwitek: These callouts render as small roman numbers (`i`, `ii`, etc.) but in the json snippet above they are  arabic numbers (`<1>`, `<2>`, etc.). This may make it harder for the user to match.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+<5> The prediction. The field name is suffixed with `_prediction`.
+<6> Indicates that this document was not used in the training set.
+
+The example above shows that the analysis has predicted the probability of all 
+the possible classes. In this case, there are two classes: `true` and `false`. 
+The class names along with the probability of the given classes are displayed in 
+the `top_classes` object. The most probable class is the prediction. In the 
+example above, `false` has a `class_probability` of 0.93 while `true` has only 
+0.06, so the prediction will be `false` which coincides with the ground truth 
+contained by the `FlightDelay` field. The class probability values help you to
```
przemekwitek: Drop "to"?

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+the possible classes. In this case, there are two classes: `true` and `false`. 
+The class names along with the probability of the given classes are displayed in 
+the `top_classes` object. The most probable class is the prediction. In the 
+example above, `false` has a `class_probability` of 0.93 while `true` has only 
+0.06, so the prediction will be `false` which coincides with the ground truth 
+contained by the `FlightDelay` field. The class probability values help you to 
+understand how sure the model is about the prediction. The higher number means 
+that the model is more confident.
+
+If a document doesn't contain a prediction field, then it is excluded from the
```
przemekwitek: "If a document doesn’t contain a prediction field, then it is excluded from the analysis."

Please clarify what do you mean by a "prediction field"? If you mean "dependent variable" then the sentence is not true.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+     "metrics": {  
+       "multiclass_confusion_matrix" : {}
+     }
+   }
+ }
+}
+--------------------------------------------------
+// TEST[skip:TBD]
+
+<1> The destination index which is the output of the analysis job.
```
przemekwitek: Similar issue with callouts (roman vs arabic).

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+    "term": {
+      "ml.is_training": {
+        "value": false  <1>
+      }
+    }
+  },
+ "evaluation": {
+   "classification": {
+     "actual_field": "FlightDelay",
+     "predicted_field": "ml.FlightDelay_prediction",
```
przemekwitek: Did you run these examples on ES? I think `predicted_field` should be `ml.FlightDelay_prediction.keyword` but please verify that.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,484 @@
...
+ }
+}
+--------------------------------------------------
+// TEST[skip:TBD]
+
+<1> By only evaluating the data that was not used in training, we can 
+calculate the generalization error which shows the algorithm accuracy in making 
+predictions for previously unseen data.
+
+The API call returns a similar response:
```
przemekwitek: Similar to what? You didn't include the response for the evaluation request above.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
szabosteve: I did run these examples in ES. As far as I know, you don't need to put `.keyword` after the name of the prediction field anymore, this issue has been fixed recently.
Screenshot 2019-11-20 at 10 00 49

przemekwitek: Ah, right. That was Dimitris' fix. Thanks for checking!

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
szabosteve: I didn't include the response of the evaluation request above as the two responses are almost identical. The structure is the same for both of them, so it is enough to highlight the differences in the requests and explain the response only once.

szabosteve:
Preview available here: http://stack-docs_690.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/flightdata-classification.html

Winterflower:
We can close this issue from the Data team side.

Winterflower:
Really sorry everyone! I was in the wrong tab and accidentally closed the wrong PR!

blaklaybul:
Looks great! I noticed that formatting is off for the last code block:

image

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,490 @@
...
+--------------------------------------------------
+PUT _ml/data_frame/analytics/model-delayed-flights
+{
+  "source": {
+    "index": [
+      "kibana_sample_data_flights"  <1>
+    ],
+    "query": {  <2>
+      "range": {
+        "DistanceKilometers": {
```
Winterflower: I need to double check this - I think I left out this query in the original classification gist, because it would not make sense to exclude flights with DistanceKilometers 0 - after all any flight that is cancelled, would probably have DistanceKilometers set to 0 (although, in this dataset one can't be too sure, since it's synthetic and a bit illoligal in places)

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,490 @@
...
+    }
+  },
+  "dest": {
+    "index": "df-flight-delayed",  <3>
+    "results_field": "ml" 
+  },
+  "analysis": {
+    "classification": {
+      "dependent_variable": "FlightDelay",  <4>
+      "training_percent": 10  <5>
```
Winterflower: It looks like perhaps the examples in this PR were taken from an older version of the Classification gist (https://gist.github.com/Winterflower/e2f226fd615a556bc1ddf9745d68df15) - in the updated version, I've bumped the training_percent to 90 to match the regression example. 
Screen Shot 2019-11-22 at 3 12 41 PM

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,490 @@
...
+      "FlightDelayType"
+    ]
+  },
+  "model_memory_limit": "100mb" <7>
+}
+--------------------------------------------------
+// TEST[skip:setup kibana sample data]
+
+<1> The source index to analyze.
+<2> This query removes erroneous data from the analysis to improve its quality.
```
Winterflower: Section [2] should be removed since we don't have the query in the example.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,490 @@
...
+{
+  "id" : "model-delayed-flights",
+  "source" : {
+    "index" : [
+      "kibana_sample_data_flights"
+    ],
+    "query" : {
+      "range" : {
+        "DistanceKilometers" : {
+          "gt" : 0
```
Winterflower: The console result should be updated, but I see that I haven't included the console response in the classification gist. I'm going to fix that.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,490 @@
...
+            "is_training" : false <7>
+          }
+----
+<1> The `dependent_variable` with the ground truth value. This is what we are 
+trying to predict with the {classanalysis}.
+<2> The top classes object contains the predicted classes with the highest 
+probability.
+<3> The probability that `FlightDelay` is false (the larger probability of the 
+two, hence this is the predicted class). This is a value between 0 and 1. The 
+higher the number the higher the probability that the datapoint belongs to the
```
Winterflower: This is suuuper, super nitpicky, but I would add a comma here to clarify the flow of the sentence a bit.
`The higher the number, the higher the probability that the datapoint belongs ...`

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
szabosteve: @Winterflower I used a smaller number because I wanted to capture the fact you mentioned on the ML weekly about even a smaller training percent leads to fairly great results. I figured it is clearer if the example represents this somehow. Would it be better to change back to 90%?

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,490 @@
...
+above, `false` has a `class_probability` of 0.93 while `true` has only 0.06, so 
+the prediction will be `false` which coincides with the ground truth contained 
+by the `FlightDelay` field. The class probability values help you understand how 
+sure the model is about the prediction. The higher number means that the model 
+is more confident.
+
+If a document doesn't contain a prediction field, then it is excluded from the 
+analysis. In order to be analyzed, a document must contain at least one field 
+with a supported data type (`numeric`, `boolean`, `text`, `keyword` or `ip`) and 
+must not contain arrays with more than one item.
```
Winterflower: I am a bit torn about what to do here, because in my opinion having different exclude behaviour for a document with one element vs doc with multiple elements in the same field is a bit confusing and I'd like to add a clarifying point for this somewhere in the docs. It might not necessarily be here in this example, though. 
I need to think about this a bit and will ping @dimitris-athanasiou for a further conversation.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,490 @@
...
+       "multiclass_confusion_matrix" : {}
+     }
+   }
+ }
+}
+--------------------------------------------------
+// TEST[skip:TBD]
+
+<1> By only evaluating the data that was not used in training, we can 
+calculate the generalization error which shows the algorithm accuracy in making
```
Winterflower: I would use the word `model` instead of the word `algorithm` here, because `model` is the one usually associated with a particular accuracy. 

`which shows the model's accuracy in making 
predictions for previously unseen data.`

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
szabosteve: You're right! Thanks!

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
+[role="xpack"]
+[testenv="platinum"]
+[[flightdata-classification]]
+=== Predicting delaying flights with {classanalysis}
```
lcawl:
```suggestion
=== Predicting delayed flights with {classanalysis}
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+
+Let's try to predict whether a flight will be delayed or not by using the 
+{kibana-ref}/add-sample-data.html[sample flight data]. We use this data to learn
```
lcawl: > We use this data to learn a classification model....

In my opinion, it might be too soon to talk about the model.  I'd suggest something more high-level like this:

For example, we want to be able to use information such as weather conditions, carrier, flight distance, origin, or destination to predict flight delays.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+[[flightdata-classification]]
+=== Predicting delaying flights with {classanalysis}
+
+Let's try to predict whether a flight will be delayed or not by using the 
+{kibana-ref}/add-sample-data.html[sample flight data]. We use this data to learn 
+a {classification} model that allows us to make the prediction based on 
+information such as weather conditions and location of the destination and 
+origin, flight distance and carrier.
+
+This dataset consists of sample data that has been manually created and it
```
lcawl: We can likely shorten this.  Not sure if "made up"/"fabricated"/etc is the right word, but here's a suggestion:
```suggestion
This sample dataset is made up and it 
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+Let's try to predict whether a flight will be delayed or not by using the 
+{kibana-ref}/add-sample-data.html[sample flight data]. We use this data to learn 
+a {classification} model that allows us to make the prediction based on 
+information such as weather conditions and location of the destination and 
+origin, flight distance and carrier.
+
+This dataset consists of sample data that has been manually created and it 
+contains inconsistencies. For example, a flight can be both delayed and 
+canceled. Nonetheless, the use case is relevant and it is a good example of how 
+the quality of input data will affect the quality of results.
```
lcawl:
```suggestion
the quality of the input data affects the quality of the results.
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+
+We create a {dfanalytics-job} to predict if a given flight is delayed. This
```
lcawl: > We create a data frame analytics job... on the field FlightDelay.

In my opinion, these first two sentences don't fit in with this paragraph.  I'd suggest moving them after the example of the document (see subsequent comment).

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+    "OriginAirportID": "ADL",
+    "OriginRegion": "SE-BD",
+    "DestCityName": "Tokoname",
+    "FlightTimeHour": 7.577903786991782,
+    "FlightDelayMin": 0
+  }
+}
+```
+
+{classification-cap} is a supervised machine learning analysis and therefore
```
lcawl:
```suggestion
Notice that each document contains a `FlightDelay` field with a boolean value. {classification-cap} is a supervised machine learning analysis and therefore 
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+    "DestCityName": "Tokoname",
+    "FlightTimeHour": 7.577903786991782,
+    "FlightDelayMin": 0
+  }
+}
+```
+
+{classification-cap} is a supervised machine learning analysis and therefore 
+needs to train on data that contains the ground truth, known as the 
+`dependent_variable`. In this example, the ground truth is available in each
```
lcawl: I think at this point we're referring to the concept of dependent variables, not to a specific entity:
```suggestion
_dependent variable_. In this example, the ground truth is available in each 
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+}
+```
+
+{classification-cap} is a supervised machine learning analysis and therefore 
+needs to train on data that contains the ground truth, known as the 
+`dependent_variable`. In this example, the ground truth is available in each 
+document as the actual value of `FlightDelay`.
+
+If your source data consists of some documents that contain a 
+`dependent_variable` and some that do not, the model is trained on the
```
lcawl: Ditto re concept vs entity:
```suggestion
dependent variable and some that do not, the model is trained on the 
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+```
+
+{classification-cap} is a supervised machine learning analysis and therefore 
+needs to train on data that contains the ground truth, known as the 
+`dependent_variable`. In this example, the ground truth is available in each 
+document as the actual value of `FlightDelay`.
+
+If your source data consists of some documents that contain a 
+`dependent_variable` and some that do not, the model is trained on the 
+`training_percent` of the documents that contain ground truth. However,
```
lcawl: I think we need more explanation of this parameter, for example:
```suggestion
subset of documents that contain ground truth (as defined in the `training_percent` parameter). However, 
```

Does this mean that if you don't specify this parameter, we'll still try (and fail) to use documents that don't contain the dependent variable?

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+document as the actual value of `FlightDelay`.
+
+If your source data consists of some documents that contain a 
+`dependent_variable` and some that do not, the model is trained on the 
+`training_percent` of the documents that contain ground truth. However, 
+predictions are made against all of the data. The current implementation of 
+{classanalysis} supports a single batch analysis for both training and 
+predictions.
+
+
```
lcawl: Optionally, re-add the sentence I suggested removing (above) here:
```suggestion
To predict whether a specific flight is delayed:
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+PUT _ml/data_frame/analytics/model-delayed-flights
+{
+  "source": {
+    "index": [
+      "kibana_sample_data_flights"  <1>
+    ]
+  },
+  "dest": {
+    "index": "df-flight-delayed",  <2>
+    "results_field": "ml"
```
lcawl: I think it's worth having a footnote for the results_field too.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+
+{classification-cap} is a supervised machine learning analysis and therefore 
+needs to train on data that contains the ground truth, known as the 
+`dependent_variable`. In this example, the ground truth is available in each 
+document as the actual value of `FlightDelay`.
+
+If your source data consists of some documents that contain a 
+`dependent_variable` and some that do not, the model is trained on the 
+`training_percent` of the documents that contain ground truth. However, 
+predictions are made against all of the data. The current implementation of
```
lcawl:
```suggestion
Predictions are made against all of the data. The current implementation of 
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+<1> The source index to analyze.
+<2> The index that will contain the results of the analysis; it will consist of 
+a copy of the source index data where each document is annotated with the 
+results.
+<3> Specifies the variable we want to predict with the {classanalysis}.
+<4> Specifies the proportion of data that is used for training the model. In 
+this example we use 10% of the source data. The training data is randomly 
+selected and provide predictions from the remaining 90%. These percentages are 
+approximate. For most of the datasets, even a small proportion of the training 
+data (2-5%) results in good predictions.
```
lcawl: > For most datasets, even a small proportion of the training data...

Can we clarify why you'd consider specifying this?  Is it so that it's faster?  Or is it so that you can use the rest as testing data?

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+    }
+  },
+  "dest" : {
+    "index" : "df-flight-delayed",
+    "results_field" : "ml"
+  },
+  "analysis" : {
+    "regression" : {
+      "dependent_variable" : "FlightDelayMin",
+      "prediction_field_name" : "FlightDelayMin_prediction",
```
lcawl: I think this prediction_field_name is the only interesting output field.  Should we draw attention to it? If not, maybe this output is unnecessary and can be omitted.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+  "allow_lazy_start" : false
+}
+--------------------------------------------------
+--
+
+. Start the job.
++
+--
+Use the {ref}/start-dfanalytics.html[start {dfanalytics-jobs}] API to start the 
+job. It will stop automatically when the analysis is complete, you don't need to
```
lcawl: Removing future tense:
```suggestion
job. It stops automatically when the analysis is complete; you don't need to 
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+[source,console]
+--------------------------------------------------
+POST _ml/data_frame/analytics/model-delayed-flights/_start
+--------------------------------------------------
+// TEST[skip:TBD]
+
+
+The job takes a few minutes to run. Runtime depends on the local hardware and 
+also on the number of documents and fields that are analyzed. The more fields 
+and documents, the longer the job to run.
```
lcawl:
```suggestion
and documents, the longer the job runs.
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+          "progress_percent" : 100
+        }
+      ]
+    }
+  ]
+}
+----  
+
+The job has four phases. When all the phases have completed, the job 
+state becomes `stopped` and the results are ready to view and evaluate.
```
lcawl: This made me wonder whether there might be other reasons why the state would be "stopped" (e.g. if you stop it manually before it's complete), so I suggest omitting that detail:
```suggestion
stops and the results are ready to view and evaluate.
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+--
+
+
+[[flightdata-classification-results]]
+==== Viewing results
```
lcawl: As is, this makes me wonder "results of what?".  If possible, I suggest clarifying. For example:
```suggestion
==== Viewing training results
```

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+
```
lcawl: I'd suggest adding an introductory paragraph at the top of this section.  For example, something about how you should now have a new index that contains a copy of your source data with predictions for your dependent variable.  That way you don't have to find all that interesting information down in the footnotes.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+possible classes. In this case, there are two classes: `true` and `false`. The 
+class names along with the probability of the given classes are displayed in the 
+`top_classes` object. The most probable class is the prediction. In the example 
+above, `false` has a `class_probability` of 0.93 while `true` has only 0.06, so 
+the prediction will be `false` which coincides with the ground truth contained 
+by the `FlightDelay` field. The class probability values help you understand how 
+sure the model is about the prediction. The higher number means that the model 
+is more confident.
+
+If a document doesn't contain a prediction field, then it is excluded from the
```
lcawl: > If a document doesn't contain a prediction field, then it is excluded from the 

This is the first time we've referred to a "prediction field", so it gave me pause about what it was referring to. I'm assuming it's referring to the existence of a field that contains the predicted value.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+class names along with the probability of the given classes are displayed in the 
+`top_classes` object. The most probable class is the prediction. In the example 
+above, `false` has a `class_probability` of 0.93 while `true` has only 0.06, so 
+the prediction will be `false` which coincides with the ground truth contained 
+by the `FlightDelay` field. The class probability values help you understand how 
+sure the model is about the prediction. The higher number means that the model 
+is more confident.
+
+If a document doesn't contain a prediction field, then it is excluded from the 
+analysis. In order to be analyzed, a document must contain at least one field
```
lcawl: > In order to be analyzed...

Sorry if I'm missing the point, but I thought at this point the analysis was already complete. Are we saying that if it doesn't have a predicted value it's because it *was* excluded from the analysis (because it contained an array or an unsupported data type, etc)?

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+       "multiclass_confusion_matrix" : {}
+     }
+   }
+ }
+}
+--------------------------------------------------
+// TEST[skip:TBD]
+
+<1> The destination index which is the output of the analysis job.
+<2> We calculate the training error by only evaluating the training data.
```
lcawl: > We calculate the training error by only evaluating the training data...

I think this is important enough to be moved out of the footnote.  For example, noting that in previous steps we saw that the new index contained a field that indicated which documents were used as training data, which we can now use to calculate the training error.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+     "metrics": {  
+       "multiclass_confusion_matrix" : {}
+     }
+   }
+ }
+}
+--------------------------------------------------
+// TEST[skip:TBD]
+
+<1> By only evaluating the data that was not used in training, we can
```
lcawl: > By only evaluating the data... 

This feels like a repetition of the lead-in sentence.  I think you could just leave the footnote as something like "Evaluate only the documents that are not training data"

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,477 @@
...
+}
+--------------------------------------------------
+<1> The name of the actual class. In this example, there are two actual classes: 
+`true` and `false`.
+<2> The number of documents in the dataset that belong to the actual class.
+<3> The name of the predicted class.
+<4> The number of documents belong to the actual class that are labelled as the 
+predicted class. 
+
+The returned confusion matrix shows us how many datapoints were classified
```
lcawl: > The returned confusion matrix shows... (actual class does not match predicted class).

I'd suggest moving this first sentence so that it introduces the response example. Otherwise, I have to wait til I get here to understand what I'm looking at.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
szabosteve: Thanks! I removed it!

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
szabosteve: Yes, I clarified this. It is important when you need to manually label your dataset with the ground truth.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
szabosteve: Thank you! It was a copy/paste error.

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -0,0 +1,442 @@
...
+<2> The number of documents in the dataset that belong to the actual class.
+<3> The name of the predicted class.
+<4> The number of documents belong to the actual class that are labelled as the 
+predicted class. 
+
+There are 8778 documents in the testing proportion of the dataset that have the 
+`false` class. The model labelled 7509 documents (out of 8778) correctly as 
+`false` and 1269 documents as `true` while those are actually `false`. There are 
+2939 documents in the testing data that have the `true` class. 1213 of them are 
+predicted as `false` and 1726 are predicted correctly as `true`.
```
sophiec20: Because it is only a binary classification example, I think it is worth additionally representing this data as tabular confusion matrix not just in words. 

Also, the sample data changes when you load it.. Please check if the values change or if it is just the timestamps. We do need the check this, because running the same analysis against different instances of Kibana where the flight data has been generated differently may result in different answers. In this case, we should qualify the results may vary.

PR #704, opened 2019-11-26
[DOCS] Expands DFA limitations section with runtime paragraph

szabosteve:
This PR adds a new limitation item to the data frame analytics limitation section about the possible runtime of the analytics jobs.

### DIFF
--- a/docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
+++ b/docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
@@ -41,6 +41,17 @@ dedicated for {ml} processes. For general {ml} settings, see
 {ref}/ml-settings.html[{ml-cap} settings in {es}].
 
 [float]
+[[dfa-time-limitations]]
+=== {dfanalytics-jobs-cap} runtime may be vary
+
+The runtime of the {dfanalytics-jobs} depends on numerous factors, such as the 
+size of the dataset, the analytics type, the number of fields that are included 
+in the analysis, the used hyperparameters, and so on. For this reason, a 
+general runtime value that applies to all or most of the situations does not 
+exist. The runtime of a {dfanalytics-job} may take from a couple of minutes up 
+to tens of hours in extreme cases.
+
+[float]
 [[dfa-missing-fields-limitations]]
 === Documents with missing values in analyzed fields are skipped
 
--- a/docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
+++ b/docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
@@ -45,11 +45,20 @@ dedicated for {ml} processes. For general {ml} settings, see
 === {dfanalytics-jobs-cap} runtime may be vary
 
 The runtime of the {dfanalytics-jobs} depends on numerous factors, such as the 
-size of the dataset, the analytics type, the number of fields that are included 
-in the analysis, the used hyperparameters, and so on. For this reason, a 
-general runtime value that applies to all or most of the situations does not 
-exist. The runtime of a {dfanalytics-job} may take from a couple of minutes up 
-to tens of hours in extreme cases.
+number of data points in the dataset, the analytics type, the number of fields 
+that are included in the analysis, the supplied hyperparameters, the type of 
+analyzed fields and so on. For example, running an analysis on a dataset with 
+many numerical fields will take longer than running an analysis on a dataset 
+that contains mainly categorical fields. Hyperparameters specified by the user 
+also lower the runtime. For this reason, a general runtime value that applies to 
+all or most of the situations does not exist. The runtime of a {dfanalytics-job} 
+may take from a couple of minutes up to 35 hours in extreme cases.
+
+The runtime increases with the increasing number of analyzed fields in a nearly 
+linear fashion. For datasets of more than 100 000 points, we recommend to start 
+with a low training percent and run a few {dfanalytics-jobs} to see how the 
+runtime scales with the increased number of data points and how the quality of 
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
```diff
@@ -40,6 +40,17 @@ You cannot update {dfanalytics-cap} configurations. Instead, delete the
 dedicated for {ml} processes. For general {ml} settings, see 
 {ref}/ml-settings.html[{ml-cap} settings in {es}].
 
+[float]
+[[dfa-time-limitations]]
+=== {dfanalytics-jobs-cap} runtime may be vary
```
lcawl:
```suggestion
=== {dfanalytics-jobs-cap} runtime may vary
```

docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
```diff
@@ -40,6 +40,17 @@ You cannot update {dfanalytics-cap} configurations. Instead, delete the
...
+
+The runtime of the {dfanalytics-jobs} depends on numerous factors, such as the 
+size of the dataset, the analytics type, the number of fields that are included
```
lcawl:
```suggestion
size of the dataset, the type of analytics, the number of fields that are included 
```

docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
```diff
@@ -40,6 +40,17 @@ You cannot update {dfanalytics-cap} configurations. Instead, delete the
 dedicated for {ml} processes. For general {ml} settings, see 
 {ref}/ml-settings.html[{ml-cap} settings in {es}].
 
+[float]
+[[dfa-time-limitations]]
+=== {dfanalytics-jobs-cap} runtime may be vary
+
+The runtime of the {dfanalytics-jobs} depends on numerous factors, such as the 
+size of the dataset, the analytics type, the number of fields that are included 
+in the analysis, the used hyperparameters, and so on. For this reason, a
```
lcawl: This isn't a term that is well known in my opinion, so I suggest including a link:
```suggestion
in the analysis, the {ref}/ml-dfanalytics-resources.html#ml-hyperparameter-optimization[hyperparameters] used, and so on. For this reason, a 
```

docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
```diff
@@ -40,6 +40,17 @@ You cannot update {dfanalytics-cap} configurations. Instead, delete the
 dedicated for {ml} processes. For general {ml} settings, see 
 {ref}/ml-settings.html[{ml-cap} settings in {es}].
 
+[float]
+[[dfa-time-limitations]]
+=== {dfanalytics-jobs-cap} runtime may vary
+
+The runtime of the {dfanalytics-jobs} depends on numerous factors, such as the 
+size of the dataset, the type of analytics, the number of fields that are included 
+in the analysis, the {ref}/ml-dfanalytics-resources.html#ml-hyperparameter-optimization[hyperparameters] used, and so on. For this reason, a
```
Winterflower: Here I think we should also talk about the type of the fields that are included in the analysis. For example, I think @tveasey has mentioned that running an analysis on a dataset with many numerical fields will take longer than running an analysis on a dataset that is purely composed of categorical fields. 
Also, we should mention that if any hyperparameters are supplied (meaning we do not have to run hyperparameter search), then this will lower the runtime (please confirm this @valeriy42 - thank you! :) )

docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
```diff
@@ -40,6 +40,17 @@ You cannot update {dfanalytics-cap} configurations. Instead, delete the
...
+[float]
+[[dfa-time-limitations]]
+=== {dfanalytics-jobs-cap} runtime may vary
+
+The runtime of the {dfanalytics-jobs} depends on numerous factors, such as the 
+size of the dataset, the type of analytics, the number of fields that are included 
+in the analysis, the {ref}/ml-dfanalytics-resources.html#ml-hyperparameter-optimization[hyperparameters] used, and so on. For this reason, a 
+general runtime value that applies to all or most of the situations does not 
+exist. The runtime of a {dfanalytics-job} may take from a couple of minutes up 
+to tens of hours in extreme cases.
```
Winterflower: There are three things I think we could improve in this paragraph:
1. While it's true that a general runtime does not exist, we can recommend that for datasets of greater than let's say 100 000 points, we recommend that users start with a low training percent and run a few jobs just to get a feel for how the runtime will scale with the increased number of datapoints and how their quality of results will scale with increased training percentage. 

@viglia 's experiment with the Rain in Australia dataset shows that with 140 000 datapoints, this dataset still has a runtime of less than 1 hour, so it would definitely be possible to run several training percentage experiments in a workday. 

2. In favour of transparency, I believe we should push the upper bound runtime from `ten hours` to at least 35 hours, which is something we have seen during testing. 

3. We should add a note that the general trend for runtime is to see it increase with increasing number of fields in a nearly linear fashion (provided that the number of datapoints analysed remains constant).

docs/en/stack/ml/df-analytics/dfanalytics-limitations.asciidoc
valeriy42: This is true :) The number of steps for  hyperparameter search is proportional to the number of undefined hyperparameters. Looking over some details, this means that going from from 4 hyperparameters to 3 would effectively reduce the time by 25%.

Issue #701, opened 2019-11-26
Write introductory content for the ML book


szabosteve:
We would like to turn the [Introduction](https://www.elastic.co/guide/en/elastic-stack-overview/master/introduction.html) section into something more useful.  Ideally, it would be similar to what’s been done for [Elasticsearch](https://www.elastic.co/guide/en/kibana/master/introduction.html). That content has the warm and welcoming tone of the definitive guide, which was very popular with our users.

Possible topics for the introduction:

- [x] Overview of when you’d choose to use anomaly detection vs outlier detection vs classification etc.
- [x] Ethical considerations (how to minimize reinforcing biases in ML) https://github.com/elastic/stack-docs/pull/721
- [x] Tips for preparing your data for analysis (e.g. workflow of ingesting, preparing, transforming, analyzing, evaluating)
- [x] Performance tips (e.g. sizing ML nodes, how to work with AD, DFA, transforms at scale)
- [x] Introduction to supervised learning https://github.com/elastic/stack-docs/pull/1306

PR #721, opened 2019-12-03
[DOCS] Adds piece about possible biases in ml processes

szabosteve:
This PR adds an introductory piece to the documentation about possible biases that can find their way into ML processes.

Related issue: https://github.com/elastic/stack-docs/issues/701

Preview available here: http://stack-docs_721.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/ml-biases.html


docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
+[[ml-biases]]
+=== Possible biases in {ml} processes
```
lcawl: I think we might want to reconsider this title, depending on where it lands in the book. If it's one of the introductory topics, we might want to make it less scary.  e.g. Change it to something like "Eliminating biases in {ml}" or "The ethics of {ml}" or "With great power..." (okay, I think I'm kidding on that last one)

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+
+{ml-cap} is meant to improve our everyday life in a wide palette of activities 
+across the globe. People use {ml} in science, healthcare, business,
```
lcawl: Not mandatory, but I think it's more compelling if we stick with "we" or "you" instead of "people", etc:
```suggestion
across the globe. We use {ml} in science, healthcare, business, 
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
+[[ml-biases]]
+=== Possible biases in {ml} processes
+
+{ml-cap} is meant to improve our everyday life in a wide palette of activities 
+across the globe. People use {ml} in science, healthcare, business, 
+entertainment, and many other fields. Decisions are made backed by {ml}
```
lcawl: Ditto re avoiding passive voice. Something like this?
```suggestion
entertainment, and many other fields. We make decisions based on {ml} results.
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
+[[ml-biases]]
+=== Possible biases in {ml} processes
+
+{ml-cap} is meant to improve our everyday life in a wide palette of activities 
+across the globe. People use {ml} in science, healthcare, business, 
+entertainment, and many other fields. Decisions are made backed by {ml} 
+techniques. In such a setup, it is crucial to keep in mind the possible biases
```
lcawl:
```suggestion
It is therefore crucial to keep in mind the possible biases 
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
+[[ml-biases]]
+=== Possible biases in {ml} processes
+
+{ml-cap} is meant to improve our everyday life in a wide palette of activities 
+across the globe. People use {ml} in science, healthcare, business, 
+entertainment, and many other fields. Decisions are made backed by {ml} 
+techniques. In such a setup, it is crucial to keep in mind the possible biases 
+that can corrupt results, endanger fairness, and jeopardize the success of even 
+the best-intended initiative. Some of these social biases can make their way to
```
lcawl:
```suggestion
the best-intended initiative. Some of these social biases can make their way into 
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
+[[ml-biases]]
+=== Possible biases in {ml} processes
+
+{ml-cap} is meant to improve our everyday life in a wide palette of activities 
+across the globe. People use {ml} in science, healthcare, business, 
+entertainment, and many other fields. Decisions are made backed by {ml} 
+techniques. In such a setup, it is crucial to keep in mind the possible biases 
+that can corrupt results, endanger fairness, and jeopardize the success of even 
+the best-intended initiative. Some of these social biases can make their way to 
+the {ml} analytics results more easily than we might think.
```
lcawl: IMO "analytics" is unnecessary here:
```suggestion
the {ml} results more easily than we might think.
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+
+There are various possible biases that can negatively influence the result of a 
+{ml} analysis. *These biases typically are not present in the {ml} algorithms 
+that are used*, but rather exist in the datasets themselves or in the way the 
+results are interpreted. We take a look at the two main types of biases: data 
+related ones and interpretation related ones and recommend some basic practices 
+to avoid them. However, it is still important to remember that the problems
```
lcawl: > We take a look...

I think this can be shortened to something like this, since you've already covered the two types in the previous sentence:

Let's consider some basic practices to avoid them.

or 

There are some basic practices that you can use to avoid them.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+related ones and interpretation related ones and recommend some basic practices 
+to avoid them. However, it is still important to remember that the problems 
+thematized in this piece are not solved yet. In fact, they are only a part of a 
+greater question about the ethics behind technology.
+
+
+[float]
+==== Biases in datasets
+
+Built-in biases are inserted in the data that we use to train the {ml} model.
```
lcawl: Inserted in seems to imply an intentional activity. Could we change to "exist" instead?
```suggestion
Built-in biases exist in the data that we use to train {ml} models. 
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+thematized in this piece are not solved yet. In fact, they are only a part of a 
+greater question about the ethics behind technology.
+
+
+[float]
+==== Biases in datasets
+
+Built-in biases are inserted in the data that we use to train the {ml} model. 
+For example, it could be the result of manual labeling when the dataset has been 
+annotated with the "ground truth" based on preconception or prejudice. The model
```
lcawl: > For example ... based on preconception or prejudice.

I think we should assume good intentions.  Maybe soften it like this?

> When we use supervised {ml} techniques, we annotate our data with the "ground truth", which is another opportunity to unwittingly introduce preconceptions or prejudice.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+==== Biases in datasets
+
+Built-in biases are inserted in the data that we use to train the {ml} model. 
+For example, it could be the result of manual labeling when the dataset has been 
+annotated with the "ground truth" based on preconception or prejudice. The model 
+will inherently learn the bias, therefore the analysis will be also infected, 
+while technically the algorithm works exactly as intended.
+
+It is impossible to avoid all the potential biases in the dataset. Eventually, 
+data is the representation of reality and, like any other type of
```
lcawl: Not a big deal, but I think there's many possible representations, not just one..
```suggestion
data is a representation of reality and, like any other type of 
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+
+Built-in biases are inserted in the data that we use to train the {ml} model. 
+For example, it could be the result of manual labeling when the dataset has been 
+annotated with the "ground truth" based on preconception or prejudice. The model 
+will inherently learn the bias, therefore the analysis will be also infected, 
+while technically the algorithm works exactly as intended.
+
+It is impossible to avoid all the potential biases in the dataset. Eventually, 
+data is the representation of reality and, like any other type of 
+representation, it never perfectly concurs with its subject. It is possible,
```
lcawl: > It is possible...

I think this is another place where it's important to make it personal.  For example:

However, you can and should carefully scrutinize...

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+however, to carefully scrutinize a dataset and make sure that it's not selected 
+in a predetermined way and it matches the real world as close as possible. To 
+involve people with diverse backgrounds in the processes that require humans can 
+also be a solution to minimize the chance of biases.
+
+
+[float]
+==== Biases in interpretation
+
+People who interpret the data can also be biased, for example, when they would
```
lcawl: > People who ...

This is another one I think we can make "you" or "we".  For example:

When we interpret data we can also be biased.  For example, we might want to validate ...

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+It is safer to formulate the aim of the analysis in a way that does not contain 
+the expected results, interpret the outcome of the analysis multiple times 
+with different persons, and perform tests as much as possible.
+
+
+[float]
+==== Collective responsibility
+
+The issue of biases that can affect {ml} related processes is far from solved. 
+In fact, this is a field of an ongoing dispute and there are no universal
```
lcawl: Probably more than one?
```suggestion
In fact, this is a field of ongoing disputes and there are no universal 
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+the expected results, interpret the outcome of the analysis multiple times 
+with different persons, and perform tests as much as possible.
+
+
+[float]
+==== Collective responsibility
+
+The issue of biases that can affect {ml} related processes is far from solved. 
+In fact, this is a field of an ongoing dispute and there are no universal 
+answers for the emerged problems. Because of the various factors that can
```
lcawl:
```suggestion
answers for the emerging problems. Because of the various factors that can 
```

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,62 @@
...
+[float]
+==== Collective responsibility
+
+The issue of biases that can affect {ml} related processes is far from solved. 
+In fact, this is a field of an ongoing dispute and there are no universal 
+answers for the emerged problems. Because of the various factors that can 
+influence the {ml} processes, avoiding bias-influenced analysis is a collective 
+responsibility of all the persons that participate in these processes; from the 
+designer of the dataset to the person who interprets the outcome of the analysis 
+to the ones who make decisions based on the data.
```
lcawl: This feels like it needs an uplifting "call to action" at the end.  Something along these lines?:

> Talk to your team about the real possibility of introducing and reinforcing biases at each step in the process. Make plans, share techniques, and be open to reconsidering results that seem too good to be true.

docs/en/stack/ml-biases.asciidoc
szabosteve: Sure! Great sentence, thank you!

docs/en/stack/ml-biases.asciidoc
szabosteve: @lcawl Also thank you for this! I assumed that preconception is something that works already unwittingly! It'll be more exact this way.

Winterflower:
Hi @blaklaybul - would be nice to hear your thoughts on this topic.

Winterflower:
hi everyone - 
this is a very important topic, so I would like to spend some time to carefully think/ do some more research about the framing of this in our Docs. For now, I think we should add a few more sections/clarifications

1) Clarifying what we mean by bias. This is a very commonly used word in many fields. For example, there is statistical bias, societal bias, legal bias etc and when we start our discussion in the docs, we should clarify what is the exact bias we are talking about. I'm assuming here that we are talking about making sure that machine learning systems don't have an adverse effect on individuals, which reinforces societal biases 

2) Mentioning examples of self-reinforcing feedback loops that machine learning systems can have. For example, predictive policing usecases. Suppose a certain city is using machine learning technologies to predict where most crimes are likely to occur. The dataset is likely to be skewed, for example, towards less wealthy areas. The ML system predicts that these areas are likely to have more crime, which leads to an increased surveillance and data collection, which further increases the bias in the dataset towards this area, which in turn leads the algo to predict higher probability of crime in this area etc. 

3) We should allow for the option that sometimes ML based/statistical systems should not be applied to people. We can take this more in-depth in a blog post, but examples of these are using people's facial features to predict ethnicity, race or sexual orientation etc. I don't think docs is perhaps the right place to talk about this in-depth, but imho it should be mentioned that there are cases when applying an ML system to classify people is dangerous and inappropriate. 

I'll think about this a bit more and come back,

Winterflower:
Here is a brief note about the ambiguities in the word `bias` from the Fair ML book
https://fairmlbook.org/introduction.html

```
The term bias is often used to refer to demographic disparities in algorithmic systems that are objectionable for societal reasons. We’ll avoid using this sense of the word bias in this book, since it means different things to different people. There’s a more traditional use of the term bias in statistics and machine learning. Suppose that Amazon’s estimates of delivery dates/times were consistently too early by a few hours. This would be a case of statistical bias. A statistical estimator is said to be biased if its expected or average value differs from the true value that it aims to estimate. Statistical bias is a fundamental concept in statistics, and there is a rich set of established techniques for analyzing and avoiding it.
```

blaklaybul:
I would be hesitant to call the page `Eliminating Bias` and change it to something like `Addressing Bias` or `Understanding Bias`. We wouldn't want to make the claim that our analyses are free of bias - they can't be. 

The following - https://data36.com/statistical-bias-types-explained/ - explains a number of different biases that arise in statistics. I think we could argue that all of these biases are relevant to the types of analyses our product enables and we should discuss them. 

Will think about this a bit more.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+results are interpreted. These issues are part of a greater question about the 
+ethics behind technology and there's no one solution for everyone. However, 
+there are some basic practices you can use to reduce these biases.
+
+
+[float]
+==== Biases in datasets
+
+Built-in biases exist in the data that we use to train {ml} models. For example,
+our choice of which data to collect or analyze unwittingly reflect our ideas
```
Winterflower: I would remove the word "unwittingly" here, because you are making conscious choices when collecting data and these choices reflect what you think is important. What you are perhaps not thinking is the unintended consequences of your choices or how your own background affects these choices.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+[float]
+==== Biases in datasets
+
+Built-in biases exist in the data that we use to train {ml} models. For example,
+our choice of which data to collect or analyze unwittingly reflect our ideas 
+about what is important or worthy of attention. But the bias could be also the 
+result of manual labeling when the dataset has been annotated with the "ground 
+truth", which is another opportunity to unwittingly introduce preconceptions or 
+prejudice. The model will inherently learn the bias, therefore the analysis will 
+be also infected, while technically the algorithm works exactly as intended.
```
Winterflower: I would use another word instead of "infected", because it has a very medical connotation.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+
+It is impossible to avoid all the potential biases in the dataset. Eventually, 
+data is a representation of reality and, like any other type of representation, 
+it never perfectly concurs with its subject. However, you can and should
```
Winterflower: I would say here that data can never perfectly capture all possible variables affecting a given phenomenon.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+[float]
+==== Collective responsibility
+
+The issue of biases that can affect {ml} related processes is far from solved. 
+In fact, this is a field of ongoing disputes and there are no universal answers 
+for the emerging problems. Because of the various factors that can influence the 
+{ml} processes, avoiding bias-influenced analysis is a collective responsibility 
+of all the persons that participate in these processes; from the designer of the 
+dataset to the person who interprets the outcome of the analysis to the ones who 
+make decisions based on the data. Talk to your team about the real possibility
```
Winterflower: Saying that a given problem is not suitable for ML applications should also be a valid option. Talk to your team about whether or not ML solutions are appropriate for this problem.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,65 @@
+[[ml-biases]]
+=== Eliminating biases in {ml} processes
```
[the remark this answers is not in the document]
blaklaybul: agree with this - other suggestions in addition to Camilla's "Understanding Biases", "Addressing Biases".

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+across the globe. We use {ml} in science, healthcare, business, entertainment, 
+and many other fields. We make decisions based on {ml} results. It is therefore 
+crucial to keep in mind the possible biases that can corrupt results, endanger 
+fairness, and jeopardize the success of even the best-intended initiative. Some 
+of these social biases can make their way into the {ml} results more easily than 
+we might think.
+
+There are various possible biases that can negatively influence the result of a 
+{ml} analysis. *These biases typically are not present in the {ml} algorithms 
+that are used*, but rather exist in the datasets themselves or in the way the
```
blaklaybul: we do have a concept of `bias` in algorithms - Bias are the simplifying assumptions made by a model to make the target function easier to learn. We might want to clarify that this doc is about the statistical/social biases and not about the bias in the algorithms.

docs/en/stack/ml-biases.asciidoc
blaklaybul: Not only what data is chosen to be collected, but what data is even available. This is a subtle difference, as a lot of training is done on available open datasets - namely people often rely on others to do the collection. I came across an interesting example recently - models for "repeat offenders" in the criminal justice system only train on - you guessed it - repeat offenders. Data is not collected for those who leave prison and act in accordance with the law. The problem here is both in the data collection (i.e. this data should be collected) and the data availability (in the absence of this data being collected, how can we address this).

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+data is a representation of reality and, like any other type of representation, 
+it never perfectly concurs with its subject. However, you can and should 
+carefully scrutinize your dataset, ensure it's not selected in a predetermined 
+way and it matches the real world as close as possible. Involve people with 
+diverse backgrounds in the processes that require humans, it is also a solution 
+to minimize the chance of biases.
+
+
+[float]
+==== Biases in interpretation
```
blaklaybul: It might be good to add something here about "blind interpretation." That is, these algorithms shouldn't be applied without a thorough understand on the data and its shortcomings arrived at through an analysis. Data is an artifact of so many different things addressed above, and this should be understood.

docs/en/stack/ml-biases.asciidoc
blaklaybul: Also, us. "from the designers of the dataset to the people who interpret the outcome of the analysis to the ones who make decisions based on the data and the companies that offer the platforms to make these decisions"

szabosteve:
@Winterflower @blaklaybul 
Thank you for taking the time to review this piece and make valuable suggestions.
I did minor changes on the document based on the feedback that you gave directly to the existing text. This was the easier part. As a next step, I'll deal with the conceptual part of the feedback which takes more time. I'll read the articles and book chapters that you suggested above then expand the text based on what I read. After that, I'd like to ask for another feedback round.
If you could help to add real examples to the text in the next round, that would be awesome.
Please let me know if you rather prefer another approach instead of the one that I proposed. Also, if you have further tips/suggestions, just let me know.

szabosteve:
@Winterflower @blaklaybul 

I fine-tuned the text according to your feedback above. I changed the title, made clearer what bias means in this context, extended the text with the topics you mentioned, and tried to phrase more carefully in general. However, I didn't address the sub-types of statistical biases individually, because it is beyond this piece as the main goal of this writing is simply to raise awareness of the issue itself.
Please have another look at the text when you have time and let me know what do you think. Thanks in advance!

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+
+{ml-cap} is meant to improve our everyday life in a wide palette of activities 
+across the globe. We use {ml} in science, healthcare, business, entertainment, 
+and many other fields. We make decisions based on {ml} results. It is, 
+therefore, crucial to keep in mind the possible biases that can corrupt results, 
+endanger fairness, and jeopardize the success of even the best-intended 
+initiative. Some of these biases can make their way into the {ml} results more 
+easily than we might think.
+
+Various possible biases can negatively influence the result of a {ml} analysis.
```
Winterflower: And add here something of the sort `and affect the people concerned`, otherwise it seems a little bit that the only reason you care about biases is because your ML analysis will be bad, while in reality you need to care about the people being impacted.

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+endanger fairness, and jeopardize the success of even the best-intended 
+initiative. Some of these biases can make their way into the {ml} results more 
+easily than we might think.
+
+Various possible biases can negatively influence the result of a {ml} analysis. 
+In this piece, we'll talk about statistical and social biases (without 
+addressing the numerous subtypes of them individually), and not about the 
+possible biases in the {ml} algorithms themselves. Statistical and social biases 
+can exist in the datasets used to train {ml} models or in the way the 
+predictions of these models are interpreted. These issues are part of a greater
```
Winterflower: Here, `These issues are part of a greater 
question about the ethics behind technology and there's no one solution for 
everyone.` , "there's no one for solution for everyone", because it is not specific enough to the topic under discussion (bias in ML) and strays a bit into a discussion of moral relativism (what x thinks is ok, is not what y thinks is ok, which is probably not what we want to allude to in the docs). Instead I'd rephrase like this

"These issues are part of a greater question about the ethics behind technology. While there are no quick technological fixes for removing bias from ML systems,  there are some basic practices you can use to detect and reduce these."

docs/en/stack/ml-biases.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+prejudice. The model will inherently learn the bias, therefore the analysis will 
+be also compromised, while technically the algorithm works exactly as intended. 
+
+It is impossible to avoid all the potential biases in the dataset. Eventually, 
+data is a representation of reality and, like any other type of representation, 
+it never perfectly captures all possible variables affecting a given phenomenon. 
+However, you can and should carefully scrutinize your dataset, ensure it's not 
+selected in a predetermined way and it matches the real world as close as 
+possible. Involve people with diverse backgrounds in the processes that require 
+humans, it is also a solution to minimize the chance of biases.
```
Winterflower: I would rewrite this sentence "Involve people with diverse backgrounds in the processes that require 
humans, it is also a solution to minimize the chance of biases." to say
"Involve people with diverse backgrounds in the design, construction and evaluation of ML systems."

szabosteve:
@Winterflower 
Your comments were very helpful, thank you for them. I amended the text based on your suggestions.

szabosteve:
@elasticmachine test this please

szabosteve:
I'm closing this PR due to changed priorities. The content of this PR may be reused later for another piece with a similar topic.

PR #723, opened 2019-12-04
[DOCS] Puts regression model info into a separated section

szabosteve:
This PR adds a section title to the paragraph about regression models in the regression conceptual documentation. This makes it easier to link directly to the section about the used model.

Preview: http://stack-docs_723.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/dfa-regression.html

### DIFF
--- a/docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
+++ b/docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
@@ -75,7 +75,7 @@ predictions are combined.
 
 [discrete]
 [[dfa-regression-model]]
-===== {regression-cap} models
+===== {regression-cap} algorithms
 
 The ensemble learning technique that we use in the {stack} is a type of boosting 
 called extreme gradient boost (XGboost) which combines decision trees with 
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
```diff
@@ -71,14 +71,20 @@ the {depvar} by combining the predictions from multiple base learners –
...
 usually better than the performance of each individual base learner because the 
 individual learners will make different errors. These average out when their 
-predictions are combined. The ensemble learning technique that we use in the 
-{stack} is a type of boosting called extreme gradient boost (XGboost) which 
-combines decision trees with gradient boosting methodologies.
+predictions are combined.
+
+[discrete]
+[[dfa-regression-model]]
+===== {regression-cap} models
```
lcawl: Is model the right term here?  In other places, we talk about creating models and model snapshots, so that's what I think of when I read the term. Whereas it seems to me this section is more about the algorithms that we use to create model.

```suggestion
===== {regression-cap} algorithms
```

docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
```diff
@@ -71,14 +71,20 @@ the {depvar} by combining the predictions from multiple base learners –
...
+
+The ensemble learning technique that we use in the {stack} is a type of boosting 
+called extreme gradient boost (XGboost) which combines decision trees with 
+gradient boosting methodologies.
 
 {regression-cap} works as a batch analysis. If new data comes into your index,
```
lcawl: > Regression works as a batch analysis...

I think this paragraph doesn't pertain to this section and belongs in the previous one.

szabosteve:
@lcawl I addressed your feedback and opened another PR that changes the title of the same section in the classification docs. This way the structure of the pieces will be in line.
https://github.com/elastic/stack-docs/pull/727

PR #727, opened 2019-12-05
[DOCS] Changes section title in classification conceptual docs

szabosteve:
This PR changes the title of the section about algorithms in the classification docs to be in line with the same section of the regression docs.

Related PR: https://github.com/elastic/stack-docs/pull/723

### DIFF
--- a/docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
+++ b/docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
@@ -42,7 +42,7 @@ supported in the {feature-var} fields.
 
 [discrete]
 [[dfa-classification-supervised]]
-==== Training the {classification} algorithm
+==== Training the {classification} model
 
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
```diff
@@ -53,8 +53,8 @@ learned about the relationships between the data points to classify new data.
 
 
 [discrete]
-[[dfa-classification-model]]
-===== {classification-cap} models
+[[dfa-classification-algorithm]]
+===== {classification-cap} algorithms
```
lcawl: I wonder if the earlier "Training the {classification} algorithm" is correct now.  Should it be "Training the {classification} model" instead?

szabosteve: Right! You train the model. I fixed it.

PR #750, opened 2019-12-11
[DOCS] Removes link refering to hyperparameter optimization in Limitations

szabosteve:
This PR removes a link that refers to the hyperparameter optimization section of the DFA resources page. As https://github.com/elastic/elasticsearch/pull/50021 will remove this page, the PR is failing because of the link in the Limitations section.

After the above-mentioned PR is merged, the link will be replaced with the new one.


szabosteve:
Backported to 7.x.

PR #756, opened 2019-12-13
[DOCS] Adds conceptual overview for influencers

lcawl:
The Kibana job tips (https://www.elastic.co/guide/en/elastic-stack-overview/master/create-jobs.html#job-tips), the results resources definition page (https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-results-resource.html), and the old ML tutorial (https://www.elastic.co/guide/en/elastic-stack-overview/7.1/ml-gs-multi-jobs.html) contain some information about influencers.  

This PR combines that information into a higher-level conceptual topic, however, and adds answers to recent questions about which fields can serve as influencers.

Preview: http://stack-docs_756.docs-preview.app.elstc.co/guide/en/elastic-stack-overview/master/ml-influencers.html


docs/en/stack/ml/anomaly-detection/influencers.asciidoc
```diff
@@ -0,0 +1,77 @@
...
+about someone or something that influences or contributes to anomalies in your
+data.
+
+Influencers can be any field in your data. If you use {dfeeds}, however, the
+field must exist in your {dfeed} query or aggregation; otherwise it is not
+included in the job analysis.
+////
+In the case of the {dfeed} query, influencers must exist in the results of the
+query in the same hit as the detector data.
+TBD: Why?
```
benwtrent: This is because when the {dfeed} processes the data, it does so by paging through the results. We try to use `doc_values` when possible, but those `doc_values` are tied to each search hit. Since search hits cannot themselves span multiple indices/documents then the {dfeed} cannot. 

It is possible that influencers and detector data are from different docs/indices when using aggs, both indices though must have the same "time_stamp" field configured.

lcawl: > ... both indices though must have the same "time_stamp" field configured.

Do you mean that the field has to have the same name?  Or format?

benwtrent: same name. Aggregations work by aggregating on field names, and that field name must have the same type (date in this case).

docs/en/stack/ml/anomaly-detection/influencers.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+included in the job analysis. If you use a query in your {dfeed}, there is an
+additional requirement: influencer fields must exist in the query results in the
+same hit as the detector fields. {dfeeds-cap} process data by paging through the
+query results; since search hits cannot span multiple indices or documents,
+{dfeeds} have the same limitation. 
+
+Influencers do not need to be fields that are specified in your {anomaly-job}
+detectors, though they often are. If you use aggregations in your {dfeed}, it is
+possible to use influencers that come from different indices than the detector
+fields. However, both indices must have the same timestamp field, which you
```
lcawl:
```suggestion
fields. However, both indices must have a date field with the same name, which you
```

docs/en/stack/ml/anomaly-detection/influencers.asciidoc
lcawl: Thanks, @benwtrent I've added that information in the following sentence: "However, both indices must have a date field with the same name, which you
specify in the `data_description`.`time_field` property for the {dfeed}." If that needs more tweaking, please just let me know!

docs/en/stack/ml/anomaly-detection/influencers.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+`customer_full_name.keyword`. By default, the swim lanes are sorted according to
+which entity has the maximum anomaly score values. You can click on the sections
+in the swim lane to see details about the anomalies that occurred in that time
+interval.
+
+TIP: The anomaly scores that you see in each section of the *Anomaly Explorer*
+might differ slightly. This disparity occurs because for each {anomaly-job},
+there are bucket results, influencer results, and record results. Anomaly scores
+are generated for each type of result. The anomaly timeline in {kib} uses the
+bucket-level anomaly scores. The list of top influencers uses the
```
peteharverson: Maybe too much detail than is needed, but note the 'view by' swim lane uses the influencer results if viewing by influencer entities, or bucket results if viewing by job ID. This section only mentions the Overall swimlane in the anomaly timeline (which does use bucket-level scores), the top influencers list and the anomalies section.

lcawl: Thanks, I've added a sentence about the influencer swim lane scores!

PR #757, opened 2019-12-13
[DOCS] Fixes incorrect version and filename in Docker GS examples.

debadair:

PR #759, opened 2019-12-16
Fix errors in docker with TLS instructions

colings86:


docs/en/getting-started/docker/create-certs.yml
```diff
@@ -2,7 +2,7 @@ version: '2.2'
 
 services:
   create_certs:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.4.0
+    image: docker.elastic.co/elasticsearch/elasticsearch:${VERSION}
```
colings86: Changed to use version from the `.env` file

docs/en/getting-started/docker/create-certs.yml
```diff
@@ -13,12 +13,15 @@ services:
...
 
 volumes: 
   certs:
-    driver: local  
\ No newline at end of file
+    driver: local  
+
+networks:
+  elastic:
+    driver: bridge
```
colings86: Used elastic network here to avoid creating two networks during the setup (previously this one would create `es_default`)

docs/en/getting-started/docker/elastic-docker-tls.yml
```diff
@@ -11,14 +11,13 @@ services:
       - cluster.initial_master_nodes=es01,es02,es03
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"      
-      - ELASTIC_PASSWORD=$ELASTIC_PASSWORD # <1>
```
colings86: The elastic password is not necessary here, the setup-passwords runs ok OOTB

docs/en/getting-started/docker/create-certs.yml
```diff
@@ -13,12 +13,15 @@ services:
         fi;
         chown -R 1000:0 /certs
       '
-    user: "0"
```
colings86: Don't need to run as root user here

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -151,7 +148,7 @@ auto --batch \
 -Expack.security.http.ssl.certificate=certificates/es01/es01.crt \
 -Expack.security.http.ssl.certificate_authorities=certificates/ca/ca.crt \
 -Expack.security.http.ssl.key=certificates/es01/es01.key \
---url https://localhost:9200"
+--url https://es01:9200"
```
colings86: I think it would be better to avoid localhost entirely. IT would be good to be able to get rid of `localhost` and `127.0.0.1` in the `instances.yml` too but thats not possible right now

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -179,7 +176,7 @@ endif::[]
 --
 ["source","sh"]
 ----
-docker-compose down
+docker-compose stop
```
colings86: `down` will tear down the containers so they have to be rebuilt, since we only need to restart them it would be better to use `stop`

debadair:
Incorporated these changes into https://github.com/elastic/stack-docs/pull/757.

PR #802, opened 2020-01-10
[DOCS] Add metadata to ML pages

lcawl:
This PR plays with adding Asciidoctor metadata to the machine learning content per https://asciidoctor.org/docs/user-manual/#metadata

In particular, this PR adds the page's content type (reference/concept/etc) as keyword metadata.  It results in the following, for example:

```
<head>
<meta name="keywords" content="concepts">
```

The values that we put in the "tags" and "subject" field in the https://github.com/elastic/docs/blob/master/conf.yaml, on the other hand, apply to every page in the book and are output as follows:

```
<head>
<meta name="DC.type" content="Learn/Docs/Elastic Stack/Machine learning/master">
<meta name="DC.subject" content="Machine learning">
```

Another option: "You can add content, such as custom metadata, stylesheet, and script information, to the header of the output document using docinfo ("document information" files)".  For example, we can have docinfo files that apply to all documents in the same directory. See https://asciidoctor.org/docs/user-manual/#docinfo-file

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/api-quickref.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/api-quickref.asciidoc
@@ -1,6 +1,7 @@
 [role="xpack"]
 [[ml-api-quickref]]
 == API quick reference
+:keywords: reference
 
 All {ml} {anomaly-detect} endpoints have the following base:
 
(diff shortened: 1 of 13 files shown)


lcawl:
Closing until we get further with discussions on how we're using this metadata

PR #821, opened 2020-01-27
[Docs][SIEM]Adds Detections UI permission error messages

benskelker:
Adds the UI permissions that can be displayed on the Detections page to the documentation.

[Preview](http://stack-docs_821.docs-preview.app.elstc.co/guide/en/siem/guide/master/detection-engine-overview.html#detections-permissions)

### DIFF
--- a/docs/en/siem/detection-engine-intro.asciidoc
+++ b/docs/en/siem/detection-engine-intro.asciidoc
@@ -81,20 +81,11 @@ Depending on your privileges and whether a `.siem-signals-<space name>` index
 has been created for the {kib} space, you might see an error message when you 
 try to open the *Detections* page.
 
-[NOTE]
-==============
-If you are using a local {es} deployment and cannot access the {siem-app}, make 
-sure the `xpack.security.enabled` setting in the `elasticsearch.yml` 
-configuration file is set to `true`. For more information, see 
-{ref}/settings.html[Configuring {es}] and
-{ref}/security-settings.html[Security settings in {es}].
-==============
-
 [float]
 ==== `Let’s set up your detection engine` message
 
 If you see this message, a user with these privileges must open the *Detections*
-page before you can view the page:
+page before you can view signals and rules:
 
 * The `manage_api_key` cluster privilege (see
 {ref}/security-privileges.html[Security privileges]).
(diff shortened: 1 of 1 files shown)


docs/en/siem/detection-engine-intro.asciidoc
```diff
@@ -77,25 +77,50 @@ To investigate a signal in the Timeline, click the *View in timeline* icon.
 [[detections-permissions]]
 === {kib} and index privileges required for Detections
 
-[IMPORTANT]
+Depending on your privileges and whether a `.siem-signals-<space name>` index 
+has been created for the {kib} space, you might see an error message when you 
+try to open the *Detections* page.
+
+[NOTE]
```
andrew-goldstein: Thanks for adding this note @benskelker! 🙏 
Please consider moving this note to the `Detection engine permissions required` section below, because running a local deployment without security enabled will result in the `Detection engine permissions required` message being displayed.

docs/en/siem/detection-engine-intro.asciidoc
```diff
@@ -77,25 +77,50 @@ To investigate a signal in the Timeline, click the *View in timeline* icon.
...
 ==============
-You cannot view the Detections page until a user with the `create_index` 
-privilege for the {kib} space visits the page.
+If you are using a local {es} deployment and cannot access the {siem-app}, make
```
andrew-goldstein: > and cannot access the {siem-app}

It is possible run a local (non-Elastic Cloud) instance without security enabled and still access the SIEM app, however the user will see the `Detection engine permissions` required message on the Overview page (in place of the `Signals` histogram), and the same message will appear in the Detections page.

Consider changing:
```
and cannot access the {siem-app}
```

to something like:
```
and you see the 'Detection engine permissions' required message in the Overview or Detections page
```

docs/en/siem/detection-engine-intro.asciidoc
```diff
@@ -77,25 +77,50 @@ To investigate a signal in the Timeline, click the *View in timeline* icon.
...
+configuration file is set to `true`. For more information, see 
+{ref}/settings.html[Configuring {es}] and
+{ref}/security-settings.html[Security settings in {es}].
 ==============
 
+[float]
+==== `Let’s set up your detection engine` message
+
+If you see this message, a user with these privileges must open the *Detections*
+page before you can view the page:
```
andrew-goldstein: > before you can view the page:

nit: Consider changing `before you can view the page` to something like `before you can enable detection rules and view signals`, because technically the user can always view the page; it just displays a message instead of showing the usual contents.

PR #850, opened 2020-02-09
[Docs][SIEM]Adds APM indices

benskelker:
Adds info about the `apm-*-transaction*` index pattern.


benskelker:
Hi @bmorelli25 - can you give this a quick look?
Thanks,

PR #851, opened 2020-02-09
[Docs][SIEM] Adds link to additional look-back time in rule ref section

benskelker:


benskelker:
Merging as just adds link

PR #858, opened 2020-02-09
[Docs][SIEM]Adds missing http tab and filebeat modules

benskelker:
Adds details to the Network page and Hosts page tabs, and adds missing filebeat modules.

Resolves #857

PR #859, opened 2020-02-10
[Docs][SIEM ]7.x add tab descriptions

benskelker:
Backports #858


docs/en/siem/siem-ui.asciidoc
```diff
@@ -86,15 +87,14 @@ Interactive widgets let you drill down for deeper insights:
...
-** Source Countries
-** Destination Countries
-* DNS
-* TLS
-* Anomalies
-* <<det-engine-terminology, External alerts>>
+* Flows: source and destination IP addresses and countries
+* DNS: DNS network queries
+* HTTP: received HTTP requests (HTTP requests for applications using
+{kibana-ref}/apm-getting-started.html[Elastic APM] are monitored by default)
```
bmorelli25: No worries since this is already merged, but we have the `{apm-app-ref}` so that when the APM content moves out of Kibana, we only have to update the link definition, instead of all the references that are scattered throughout the documentation.

benskelker: ok - i'll change that today. it might save us some hassle in the future...

bmorelli25: Thanks. Sorry about all the trouble 😅

PR #885, opened 2020-02-17
[DOCS] Adds section to the classification docs that helps interpret results

szabosteve:
This PR adds an explanation of `class_probability` and `class_score` to the classification docs.

Preview: http://stack-docs_885.docs-preview.app.elstc.co/guide/en/machine-learning/master/dfa-classification.html

### DIFF
--- a/docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
+++ b/docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
@@ -93,23 +93,25 @@ The value of `class_score` controls the probability at which a class label is
 assigned to a datapoint. In normal case – that you maximize the number of 
 correct labels – a class label is assigned when its predicted probability is 
 greater than 0.5. The `class_score` makes it possible to change this behavior, 
-so it can be less than or greater than 0.5. The value of `class_score` is always 
+so it can be less than or greater than 0.5. For example, suppose our two classes 
+are denoted `class 0` and `class 1`, then the value of `class_score` is always 
 non-negative and its definition is:
 
 ```
-scoreClass0 = 0.5 / (1.0 - probabilityAtWhichToAssignClassOne) * probabilityClass0
-scoreClass1 = 0.5 / probabilityAtWhichToAssignClassOne * probabilityClass1
+class_score(class 0) = 0.5 / (1.0 - k) * probability(class 0)
+class_score(class 1) = 0.5 / k * probability(class 1)
 ```
 
-`probabilityAtWhichToAssignClassOne` **is chosen to maximise the minimum 
-recall**. It is useful for example in case of highly imbalanced data. If 
-`class 0` is much more frequent in the training data than `class 1`, then it 
-can mean that you achieve the best accuracy by assigning `class 0` to every 
-datapoint. This is equivalent to zero recall for `class 1`. Instead of this 
-behavior, the default scheme of the {stack} {classanalysis} is to choose 
-`probabilityAtWhichToAssignClassOne < 0.5` and accept a higher rate of actual 
-`class 0` predicted `class 1` errors, or in other words, a slight degradation of 
-the overall accuracy.
+Here, `k` is a positive constant less than one. It represents the predicted 
+probability of `class 1` for a datapoint at which to label it `class 1` and is 
+chosen to maximise the minimum recall of any class. This is useful for example 
+in case of highly imbalanced data. If `class 0` is much more frequent in the 
+training data than `class 1`, then it can mean that you achieve the best 
+accuracy by assigning `class 0` to every datapoint. This is equivalent to zero 
+recall for `class 1`. Instead of this behavior, the default scheme of the 
+{stack} {classanalysis} is to choose `k < 0.5` and accept a higher rate of 
+actual `class 0` predicted `class 1` errors, or in other words, a slight 
+degradation of the overall accuracy.
 
 
 [discrete]
--- a/docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
+++ b/docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
@@ -65,9 +65,56 @@ boosted tree regression model which combines multiple weak models into a
 composite one. We use decision trees to learn to predict the probability that 
 a data point belongs to a certain class.
 
+
+[discrete]
+[[dfa-classification-interpret]]
+==== Interpreting {classification} results
+
+The following sections help you understand and interpret the results of a 
+{classanalysis}.
+
+[discrete]
+[[dfa-classification-class-probability]]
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
```diff
@@ -65,9 +65,56 @@ boosted tree regression model which combines multiple weak models into a
...
+==== Interpreting {classification} results
+
+The following sections help you understand and interpret the results of a 
+{classanalysis}.
+
+[discrete]
+[[dfa-classification-class-probability]]
+===== `class_probability`
+
+The value of `class_probability` shows how likely that a given datapoint belongs
```
tveasey:
```suggestion
The value of `class_probability` shows how likely it is that a given datapoint belongs 
```

docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
```diff
@@ -65,9 +65,56 @@ boosted tree regression model which combines multiple weak models into a
...
+non-negative and its definition is:
+
+```
+scoreClass0 = 0.5 / (1.0 - probabilityAtWhichToAssignClassOne) * probabilityClass0
+scoreClass1 = 0.5 / probabilityAtWhichToAssignClassOne * probabilityClass1
+```
+
+`probabilityAtWhichToAssignClassOne` **is chosen to maximise the minimum 
+recall**. It is useful for example in case of highly imbalanced data. If 
+`class 0` is much more frequent in the training data than `class 1`, then it
```
tveasey: This need not be the case, so I'd change

"it means that you can achieve the best accuracy when every datapoint is considered 
as `class 0`"

to

"it can mean that you achieve the best accuracy by assigning `class 0` to every datapoint"

docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
```diff
@@ -65,9 +65,56 @@ boosted tree regression model which combines multiple weak models into a
...
+
+[discrete]
+[[dfa-classification-class-score]]
+===== `class_score`
+
+The value of `class_score` controls the probability at which a class label is 
+assigned to a datapoint. In normal case – that you maximize the number of 
+correct labels – a class label is assigned when its predicted probability is 
+greater than 0.5. The `class_score` makes it possible to change this behavior, 
+so it can be less than or greater than 0.5. The value of `class_score` is always
```
tveasey: Thinking a bit more about this I think we need to somehow motivate class 0 and class 1 a bit better. How about tweaking the following slightly along the following lines:

For example, suppose our two classes are denoted class 0 and class 1, then the value of `class_score` is always
non-negative and its definition is:

```
score(class 0) = 0.5 / (1.0 - k) * probability(class 0)
score(class 1) = 0.5 / k * probability(class 1)
```

Here, `k` is a positive constant less than one. It represents the predicted probability of class 1 for a datapoint at which to label it class 1 and is chosen to maximise the minimum recall of any class.

docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
```diff
@@ -65,9 +65,58 @@ boosted tree regression model which combines multiple weak models into a
...
+{classanalysis}.
+
+[discrete]
+[[dfa-classification-class-probability]]
+===== `class_probability`
+
+The value of `class_probability` shows how likely it is that a given datapoint 
+belongs to a certain class. It is a value between 0 and 1. The higher the 
+number, the higher the probability that the data point belongs to the named 
+class. For further details, see the
```
lcawl: I think we need to qualify this link a bit more, since the example focuses on viewing the results in Kibana, which I don't think shows this class_probability value by default.  For example:
```suggestion
class. This information is stored in the `top_classes` array for each document in your destination index. See the
```

docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
```diff
@@ -65,9 +65,58 @@ boosted tree regression model which combines multiple weak models into a
...
+[discrete]
+[[dfa-classification-interpret]]
+==== Interpreting {classification} results
+
+The following sections help you understand and interpret the results of a 
+{classanalysis}.
+
+[discrete]
+[[dfa-classification-class-probability]]
+===== `class_probability`
```
lcawl: This information already appears in the example (flightdata-classification.html#flightdata-classification-results) and in my opinion that's a better place for it, unless we end up creating a separate page that dives into how to interpret classification results.

docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
```diff
@@ -65,9 +65,58 @@ boosted tree regression model which combines multiple weak models into a
...
+belongs to a certain class. It is a value between 0 and 1. The higher the 
+number, the higher the probability that the data point belongs to the named 
+class. For further details, see the
+{ml-docs}/flightdata-classification.html#flightdata-classification-results[Viewing {classification} results]
+section in the {classification} example.
+
+
+[discrete]
+[[dfa-classification-class-score]]
+===== `class_score`
```
lcawl: It is unclear where to find this score.  I don't see it in the results example in http://stack-docs_885.docs-preview.app.elstc.co/guide/en/machine-learning/master/flightdata-classification.html#flightdata-classification-results

docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
szabosteve: @lcawl 
I think it is a good solution to collect this kind of info and publish it in one place.  You are right that this info appears on the example page and it is needed there for the full context. I also would like to add here, so you'll have a basic understanding of how to interpret results even if you don't read the examples. WDYT?

docs/en/stack/ml/df-analytics/dfa-classification.asciidoc
szabosteve: AFAIK, `class_score` has become exposed recently. I'll update the API examples in a separate PR.

Winterflower:
Hey everyone, I've been reading the text in this PR and at the moment I'm not sure someone just starting out with the product would have an easy time understanding what's being said here. For example, we don't define recall anywhere. To the layman it has a different meaning than in machine learning, so for many people in Ops, Security etc. `minimum recall` might not actually say anything useful. I'll think about how to phrase this a bit today and will post some feedback.

szabosteve:
We decided to merge the PR as is now. We'll add the `minimum recall` definition to the glossary than put a link in the text that points to the glossary item.

szabosteve:
**Backport commits:**
7.x: 0746f43
7.6: 00e9884

PR #888, opened 2020-02-18
[DOCS] Adds influencer to Stack glossary

lcawl:
This PR adds "influencer" to the Elastic Stack glossary (https://www.elastic.co/guide/en/elastic-stack-glossary/current/terms.html). It also adds an additional sentence to the influencer overview in the Machine learning documentation (https://www.elastic.co/guide/en/machine-learning/current/ml-influencers.html). 

It also wraps some existing glossary entries to align with typical line lengths.

### DIFF
--- a/docs/en/glossary/glossary.asciidoc
+++ b/docs/en/glossary/glossary.asciidoc
@@ -373,6 +373,15 @@ order to index <<glossary-event,event>> data.
 +
 //Source: Logstash
 endif::logstash-terms[]
+ifdef::xpack-terms[]
+[[glossary-influencer]] influencer ::
+
+Influencers are fields that affect whether a bucket is anomalous in an
+<<glossary-anomaly-detection-job,{anomaly-job}>>. For more information, see
+{ml-docs}/ml-influencers.html[Influencers].
++
+//Source: X-Pack
+endif::xpack-terms[]
 ifdef::logstash-terms[]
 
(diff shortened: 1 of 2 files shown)


docs/en/glossary/glossary.asciidoc
```diff
@@ -373,6 +373,15 @@ order to index <<glossary-event,event>> data.
 +
 //Source: Logstash
 endif::logstash-terms[]
+ifdef::xpack-terms[]
+[[glossary-influencer]] influencer ::
+
+Influencers are fields that affect whether a bucket is anomalous in an
```
Winterflower: Hi Lisa!
I've checked how we talked about influencers in some past blog posts and I think it would be better to revise this glossary entry to be "Influencers are entities that may have contributed to an anomaly in a given bucket." to be more in line with our previous writing on this topic. 
Do you think we should mention things like "These entities will, in practice, be values of fields in the documents of the analyzed dataset"? Or is this too much for a glossary?

lcawl: Thanks @Winterflower I've changed it to this:
> Influencers are entities that might have contributed to an anomaly in a specific
bucket in an {anomaly-job}.

Winterflower: Thank you @lcawl - it looks good!

PR #895, opened 2020-02-20
[DOCS][Endpoint]: Add temp Endpoint Docs Home

DonNateR:


docs/en/endpoint/index.asciidoc
```diff
@@ -0,0 +1,6 @@
+[[elastic-endpoint]]
+= Elastic Endpoint Security
+
```
bmorelli25: I would probably add your includes for versions and attributes here. They may not be necessary yet, but IMO, it'd be better to have everything set up now, so you don't have to add it later.

```
include::{asciidoc-dir}/../../shared/versions/stack/{source_branch}.asciidoc[]

include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
```

bmorelli25:
@elasticmachine, run elasticsearch-ci/docs

bmorelli25:
Oops, wrong PR :)

PR #918, opened 2020-03-03
[Docs][ML]Updates prebuilt ml jobs file links

benskelker:
Updates links in the ML SIEM prebuilt jobs section so it can be included in the SIEM docs without breaking links.

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
@@ -524,16 +524,6 @@ below contains the jobs organized by `agent.type` (Auditbeat, Packetbeat, and
 Winlogbeat).
 
 // tag::siem-jobs[]
-ifeval::[{docname} == "ootb-ml-jobs"]
-:rare-func-ref: <<ml-rare,`rare` function>> 
-:nonzero-count-func-ref: <<ml-nonzero-count,`high_non_zero_count` function>>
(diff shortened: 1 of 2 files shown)


docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -517,13 +517,23 @@ Function: `non_zero_count`.
...
 === SIEM
 
-// tag::siem-jobs[]
 These {anomaly-jobs} appear by default in the Anomaly Detection interface of the 
 {siem-guide}/machine-learning.html[SIEM app] in {kib}. They help you 
 automatically detect file system and network anomalies on your hosts. The list 
 below contains the jobs organized by `agent.type` (Auditbeat, Packetbeat, and 
 Winlogbeat).
 
+// tag::siem-jobs[]
```
lcawl: This seems likely to work fine, though for future reference I think it's fine to just make them as external links (i.e. use xxx.html instead of <<xxx>>) in the context of both books. That's what we did for links in release highlights and breaking changes (e.g. https://www.elastic.co/guide/en/elastic-stack/current/elasticsearch-highlights.html) which are re-used in both the Installation/Upgrade Guide and individual product books.

benskelker: I'll do this now as it keeps thing simple.

benskelker:
@szabosteve - can you just give this another quick look? I updated as Lisa suggested.
Thanks,

PR #921, opened 2020-03-04
[Docs][SIEM]Adds PATCH method to update rule api

benskelker:
Describes using PATCH and PUT for updating detection engine rules. 

[Preview](http://stack-docs_921.docs-preview.app.elstc.co/guide/en/siem/guide/master/rules-api-update.html)


FrankHassanabad:
Bulk updates can also do PATCH as well so this part needs the small addition to it.

Screen Shot 2020-03-04 at 9 59 00 PM

FrankHassanabad:
👍 

That looks perfect!

Screen Shot 2020-03-04 at 10 00 40 PM

PR #932, opened 2020-03-09
[Docs][ML][SIEM] Combines prebuilt job descriptions

benskelker:
Merges the SIEM ML prebuilt job information into one section.

Preview:
* http://stack-docs_932.docs-preview.app.elstc.co/guide/en/siem/guide/master/prebuilt-ml-jobs.html
* http://stack-docs_932.docs-preview.app.elstc.co/guide/en/machine-learning/master/ootb-ml-jobs-siem.html


benskelker:
@randomuserid @dcode 
This doesn't change the descriptions of the ML jobs, it single sources the section from the ML docs, and adds more info.

docs/en/siem/machine-learning.asciidoc
```diff
@@ -71,399 +71,4 @@ do not need to manually map the fields.
...
-persistence using compromised accounts. RDP attacks such as BlueKeep also tend 
-to use unusual usernames.
-+
-Beats required on hosts:
-
-* Winlogbeat (Windows)
-
-NOTE: You can only run this job when Winlogbeat is used to ship data.
\ No newline at end of file
+include::../stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc[tag=siem-jobs]
```
lcawl: We typically make cross-book inclusions relative to this intrinsic attribute (https://www.methods.co.nz/asciidoc/chunked/ch31.html):
```suggestion
include::{asciidoc-dir}/../stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc[tag=siem-jobs]
```

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -519,522 +519,613 @@ Function: `non_zero_count`.
...
+Beats required on hosts:::
 
-////
-Influencers:
+* Auditbeat (Linux)
+
++
+NOTE: You can only run this job when Auditbeat is used to ship data.
+
+{ml-docs}/ml-influencers.html[Influencers]:::
```
lcawl: Shouldn't this match the other sections and have this title and then "(influencer)" after each field?
```suggestion
Required ECS fields:::
```

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -519,522 +519,613 @@ Function: `non_zero_count`.
...
```
lcawl: Ditto re changing title and adding "(influencer)" after each field:
```suggestion
Required ECS fields:::
```

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -519,522 +519,613 @@ Function: `non_zero_count`.
...
+Beats required on hosts:::
 
-////
-Influencers:
+* Winlogbeat (Windows)
+
++
+NOTE: You can only run this job when Winlogbeat is used to ship data.
+
+{ml-docs}/ml-influencers.html[Influencers]:::
```
lcawl: Ditto
```suggestion
Required ECS fields:::
```

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -519,522 +519,613 @@ Function: `non_zero_count`.
...
```
lcawl: Ditto:
```suggestion
Required ECS fields:::
```

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -519,522 +519,613 @@ Function: `non_zero_count`.
...
+execution of unauthorized network applications.
 
-* For network activity logs where `agent.type` is `auditbeat`.
+Job details:::
+
+* Analyzes network activity logs where `agent.type` is `auditbeat` (Linux) or `winlogbeat` (Windows).
 * Models the occurrences of processes that cause network activity.
 * Detects network activity caused by processes that occur rarely compared to 
   other processes (using the {ml-docs}/ml-rare-functions.html#ml-rare[`rare` function]).
+* {ml-docs}/ml-buckets.html[Bucket span]: 15 minutes.
```
lcawl: In my opinion, these repeated links to info about bucket spans and influencers are distracting.  I'd suggest having them appear once at the top of the page.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
benskelker: @lcawl  
hmm not sure here. From our perspective, it's only interesting when users are not using beats and can manually map their data to ECS. When a job can be mapped, the influencers are always ECS fields. When a job cannot be mapped because it relies on non-ECS beat fields, I listed the influencers separately.

What do you think?

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
dcode: I agree with this. Maybe before the first job have a short description of what each of the terms mean  (single sentence summary) with the existing hyperlink for more detailed info. And maybe have a single list of the jobs at the top of the page with links to the job itself. I'm thinking like the [Logstash mutate filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-mutate.html#plugins-filters-mutate-options)

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
dcode: Are the non-influencer fields "detector" fields? If I was reading this documentation for the first time trying to understand the ML job options, it's not clear to me what these other fields are doing.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
benskelker: I think a lot of people will be using ctrl/cmd f on this page, searching for a specific job.
I don't mind though.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
lcawl: Sorry, I'm not sure I understand the difference.  These look like the same field names that exist in other "ECS fields" sections, so if they're not the same I think further clarification is required.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -519,522 +519,613 @@ Function: `non_zero_count`.
...
 // tag::siem-jobs[]
-[float]
-[[ootb-ml-jobs-siem-audit]]
-==== SIEM - Auditbeat
-
 linux_anomalous_network_activity_ecs::
++
+windows_anomalous_network_activity_ecs::
+
+Identifies Windows/Linux processes that do not usually use the network but have
```
dcode:
```suggestion
Identifies processes that do not usually use the network but have
```
This is maybe more style, but I think this makes the sentence easier to read. The OS specifiers in the names and in the required query I think clearly identify what OS the process is running on.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
dcode: The way I read the description, the `agent.type` must be as specified in these jobs, so ECS-compliant data from other sources wouldn't work. Is that accurate?

benskelker: This is the intro sentence in the SIEM docs, before including your stuff:

If you use Beats to ship your data, the ECS fields are populated and you 
do not need to manually map the fields.

Not sure, but maybe add something to your file before the include?

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
benskelker: Sure, but we'll wait for @randomuserid as the descriptions are basically his.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
benskelker: > The way I read the description, the `agent.type` must be as specified in these jobs, so ECS-compliant data from other sources wouldn't work. Is that accurate?

Don't think so. The `agent.type` would need to be mapped to the user's data shipper. The issue with some jobs is that they rely on non-compliant ECS fields, so the only way they can be run is when beats are used to ship the data.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
lcawl: @benskelker Per our chat, I think the bucket span and influencer details can be omitted from this list (and therefore the links to bucket span and influencer definitions too). One thing we might consider adding to all these pages, however, is a link to where the modules are defined: https://github.com/elastic/kibana/tree/master/x-pack/plugins/ml/server/models/data_recognizer/modules

Then folks can dig into the exact job and datafeed definitions if they so desire.  For the SIEM auditbeat ones can be found here: https://github.com/elastic/kibana/tree/master/x-pack/plugins/ml/server/models/data_recognizer/modules/siem_auditbeat/ml

lcawl: I've merged that change in https://github.com/elastic/stack-docs/pull/933. If you want help with the merge conflict, please just let me know!

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -286,254 +286,568 @@ visitor_rate_ecs::
...
 === SIEM
 
-These {anomaly-jobs} appear by default in the Anomaly Detection interface of the 
-{siem-guide}/machine-learning.html[SIEM app] in {kib}. They help you 
-automatically detect file system and network anomalies on your hosts. The list
-below contains the jobs organized by `agent.type` ({auditbeat}, {packetbeat},
-and {winlogbeat}).
+These {anomaly-jobs} appear by default in the Anomaly Detection interface of
+the {siem-guide}/machine-learning.html[SIEM app] in {kib}. They help you 
+automatically detect file system and network anomalies on your hosts.
```
benskelker: @szabosteve @lcawl 
Don't know if you want to add some info here as to why the ECS fields are listed for each job?

docs/en/siem/machine-learning.asciidoc
benskelker: @lcawl 
Can you help? I couldn't get this to resolve. Is it only true when including files from the docs repo, not when including tags.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
szabosteve: @benskelker @lcawl 
Sure, I'll add in a separate PR that when you are not using Beats, you need to map your data to the respective ECS fields.

szabosteve:
```suggestion
automatically detect file system and network anomalies on your hosts. However,
if you don't use Beats, you need to map your data to the ECS fields that are 
listed for every job.
```

szabosteve: @benskelker @lcawl 
I added a sentence about the listed ECS fields as a single comment to avoid conflicts. Please let me know if you would modify or rather omit it.

benskelker: Thanks @szabosteve

docs/en/siem/machine-learning.asciidoc
lcawl: You're right.  I've changed it just use an "ml-dir" attribute, in case that path changes in the future.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -286,254 +286,568 @@ visitor_rate_ecs::
...
+Searches for activity from users who are not normally active, which can
+indicate unauthorized changes, activity by unauthorized users, lateral
+movement, and compromised credentials.
++
+In organizations, new usernames are not often created apart from specific types 
+of system activities, such as creating new accounts for new employees. These 
+user accounts quickly become active and routine.
++
+Events from rarely used usernames can point to suspicious activity. 
+Additionally, automated Linux fleets tent to see activity from rarely used
```
lcawl:
```suggestion
Additionally, automated Linux fleets tend to see activity from rarely used 
```

docs/en/siem/machine-learning.asciidoc
```diff
@@ -66,404 +69,10 @@ Management -> Advanced Settings -> `siem:defaultAnomalyScore`.
...
-* event.code
-* agent.type 
-
-windows_rare_user_type10_remote_login:: Searches for unusual remote desktop 
-protocol (RDP) logins, which could indicate account takeover or credentialed 
-persistence using compromised accounts. RDP attacks such as BlueKeep also tend 
-to use unusual usernames.
-+
-Beats required on hosts:
+hosts. To run jobs when you are not using {beats}, you must map your data to
```
lcawl: With other ML modules, you don't even see the job suggestions unless you have the appropriate data. If that's true here as well, it might help to do something like this:
```suggestion
hosts. If you are not using {beats}, you must map your data to
```

docs/en/siem/machine-learning.asciidoc
```diff
@@ -66,404 +69,10 @@ Management -> Advanced Settings -> `siem:defaultAnomalyScore`.
...
+the ECS fields listed in each job description.
```
lcawl: Continuation of previous suggestion...
```suggestion
ECS fields in order to see these job recommendations.
```

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
```diff
@@ -286,254 +286,568 @@ visitor_rate_ecs::
...
   {ml-docs}/ml-info-functions.html#ml-info-content[`high_info_content` function]).
 
-Looks for unusual powershell scripts that may indicate execution of malware, or 
-persistence mechanisms.
+Required {beats}:::
+
+* {winlogbeat} (Windows)
+
++
+NOTE: You can only run this job when Winlogbeat is used to ship data.
```
lcawl: Attempting to clarify whether there is a failure to run (i.e. job is visible but fails) or whether you never see this job:
```suggestion
NOTE: This job is available only when you use {winlogbeat} to ship data.
```

docs/en/siem/machine-learning.asciidoc
benskelker: In SIEM, you can always see the jobs. They are marked as unavailable. Changing to:
To enable jobs when you are not using {beats},...

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs.asciidoc
benskelker: Yep - will update all the relevant notes.

Issue #948, opened 2020-03-22
[Docs]Add dedicated rule version history section


benskelker:
Users need to keep track of prebuilt detections rules and their changes:

https://github.com/elastic/kibana/pull/60825

PR #949, opened 2020-03-22
[Docs][SIEM] Adds prebuilt rule version history section

benskelker:
So users can track changes to prebuilt rules over time: https://github.com/elastic/kibana/pull/60825

[Preview](http://stack-docs_949.docs-preview.app.elstc.co/guide/en/siem/guide/master/prebuilt-rules-changelog.html)

Resolves: https://github.com/elastic/stack-docs/issues/948 

Even though there is a lot of repetition, I think we have to slice the changes by rule. Otherwise, in the future, it'll be really hard to see changes over time for a specific rule.
If anyone has any other suggestions about the structure, please let me know.


dcode:
What if the complete changelog for each rule was on the rule-specific page ([like here](http://stack-docs_949.docs-preview.app.elstc.co/guide/en/siem/guide/master/dns-activity-to-the-internet.html)), and the summary like you have [here](http://stack-docs_949.docs-preview.app.elstc.co/guide/en/siem/guide/master/prebuilt-rules-changelog.html) could be part of the release changelog to include rules that changed. Just an idea. I do like this format as well (and was unaware of the `%collapsible` feature of asciidoctor, which is awesome)

benskelker:
@dcode - Thanks. I think we need a rule-specific changelog. Otherwise, it'll be hard to find all the rule changes over time. Not adding the changes to the rule-specific pages is for convenience, so I can update the pages with a script. If you think it's necessary, I'll add the changes to individual rule pages.

Issue #1026, opened 2020-04-26
Threat hunting enhancements


benskelker:
Document threat hunting enhancements: https://github.com/elastic/kibana/pull/61207

cc: @andrew-goldstein @XavierM @MikePaquette

PR #1039, opened 2020-05-06
[Docs][SIEM]Threat hunting ehancements

benskelker:
Documents threat hunting UI interactions.

[Preview](http://stack-docs_1039.docs-preview.app.elstc.co/guide/en/siem/guide/master/siem-ui-overview.html)

Resolves: https://github.com/elastic/stack-docs/issues/1026


docs/en/siem/siem-ui.asciidoc
```diff
@@ -8,7 +8,31 @@ environment, and you can use the interactive UI to drill down into areas of
 interest.
 
 The *{kibana-ref}/kuery-query.html[{kib} Query Language (KQL)]* bar is available
-throughout the {siem-app} for searching and filtering.
+throughout the {siem-app} for searching and filtering. You can also click and
+swipe histograms, which updates the global time filter.
```
andrew-goldstein, line 12: > You can also click and swipe histograms, which updates the global time filter.

Would you be willing to consider describing this behavior ("brush selection" is the obscure technical term for it), like the following suggestion?

When histograms contain peaks or a time range containing interesting data, click on the chart, and while holding down the mouse or trackpad button, drag over the time range to select it, which updates the global time filter.

brush-selection-still

Consider including a still screenshot like the one above to illustrate the selection process.

Do we support animated gifs in docs? If so, perhaps we could include something like the following?

brush-selection-animated

docs/en/siem/siem-ui.asciidoc
```diff
@@ -8,7 +8,31 @@ environment, and you can use the interactive UI to drill down into areas of
...
+
+[role="screenshot"]
+image::images/siem-click-swipe.png[]
+
+TIP: All {siem-soln} histograms and graphs contain an **Inspect** button, so
+you can examine data sources throughout the app.
```
andrew-goldstein, line 18: > so you can examine data sources throughout the app.

Consider replacing "data sources" with "queries", e.g.:

so you can view the Elasticsearch queries that provide data for graphs throughout the app.

docs/en/siem/siem-ui.asciidoc
benskelker: Updated the UI text to describe what to do instead of how to do it, and the screenshot. I think it's ok now.
Technically, we can add gif animations. I need to check with other writers if it's ok to do so. Personally, I find them a bit distracting in docs.

PR #1047, opened 2020-05-07
[DOCS] Update snapshot retention details

lcawl:
This PR updates the "Model snapshots" page with details about how long snapshots are retained. It also includes minor improvements to other sections of that page.

Related to https://github.com/elastic/elasticsearch/issues/52150

Preview: http://stack-docs_1047.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-model-snapshots.html

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/model-snapshots.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/model-snapshots.asciidoc
@@ -14,10 +14,12 @@ approximately every 3 to 4 hours. You can change this interval
 (`background_persist_interval`) when you create or update a job.
 
 To reduce the number of snapshots consuming space on your cluster, at the end of
-each day, old snapshots are automatically deleted. The age of each snapshot is calculated relative to the timestamp of the most recent snapshot. By default, if there are
-snapshots over one day older than the most recent snapshot, only the first snapshot for each day is retained.
-As well, all snapshots over ten days old are deleted. You can change these
-retention settings (`daily_model_snapshot_retention_after_days` and
+each day, old snapshots are automatically deleted. The age of each snapshot is
+calculated relative to the timestamp of the most recent snapshot. By default, if
+there are snapshots over one day older than the newest snapshot, they are
+deleted except for the first snapshot each day. As well, all snapshots over ten
+days older than the newest snapshot are deleted. You can change these retention
+settings (`daily_model_snapshot_retention_after_days` and
 `model_snapshot_retention_days`) when you create or update a job. If you want to
 exempt a specific snapshot from this clean up, use the
 {ref}/ml-update-snapshot.html[update model snapshots API] to set `retain` to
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/anomaly-detection/model-snapshots.asciidoc
```diff
@@ -8,21 +8,28 @@ accomplished by generating models of your data.
...
+cluster. The amount of time necessary to save these snapshots is proportional to
+the size of the model in memory. By default, snapshots are captured
+approximately every 3 to 4 hours. You can change this interval
+(`background_persist_interval`) when you create or update a job.
 
-You can use the {ref}/ml-update-job.html[update {anomaly-jobs} API] to change
-the interval (`background_persist_interval`) and retention
-(`model_snapshot_retention_days`) of these snapshots.
+To reduce the number of snapshots consuming space on your cluster, at the end of
+each day, old snapshots are automatically deleted. By default, if there are
```
lcawl, line 17: Do we need to cover how age is calculated here too? For example:
```suggestion
each day, old snapshots are automatically deleted. The age of each snapshot is calculated relative to the timestamp of the most recent snapshot. By default, if there are
```

docs/en/stack/ml/anomaly-detection/model-snapshots.asciidoc
```diff
@@ -8,21 +8,28 @@ accomplished by generating models of your data.
...
+snapshots over one day old, only the first snapshot for each day is retained.
```
lcawl, line 18: Likewise, incorporating relative age:
```suggestion
snapshots over one day older than the most recent snapshot, only the first snapshot for each day is retained.
```

docs/en/stack/ml/anomaly-detection/model-snapshots.asciidoc
droberts195: Yes, we probably should, because it's natural to think of "age" as being relative to now, and that's not the case here.

lcawl: Thanks, I've implemented those changes.

docs/en/stack/ml/anomaly-detection/model-snapshots.asciidoc
```diff
@@ -8,21 +8,30 @@ accomplished by generating models of your data.
 
 To ensure resilience in the event of a system failure, snapshots of the {ml}
 model for each {anomaly-job} are saved to an internal index within the {es}
-cluster. By default, snapshots are captured approximately every 3 to 4 hours and 
-retained for one day (twenty-four hours). The amount of time necessary to
-save these snapshots is proportional to the size of the model in memory.
+cluster. The amount of time necessary to save these snapshots is proportional to
+the size of the model in memory. By default, snapshots are captured
+approximately every 3 to 4 hours. You can change this interval
+(`background_persist_interval`) when you create or update a job.
```
szabosteve, line 14: Should we link to the APIs where the users can change the interval?

PR #1099, opened 2020-05-20
Remove LS steps from the stack getting started

### LINKED ISSUE #1098: Remove Logstash from the stack getting started guide
We don't want to send users down the path of using Logstash as part of the getting started experience.

Let's remove Logstash and instead point to the getting started docs for LS.

dedemorton:
Closes #1098

### DIFF
--- a/docs/en/getting-started/get-started-stack.asciidoc
+++ b/docs/en/getting-started/get-started-stack.asciidoc
@@ -6,10 +6,14 @@ Looking for an {stack} ("ELK") guide that shows how to quickly install and confi
 * <<install-elasticsearch,{es}>>
 * <<install-kibana,{kib}>>
 * <<install-beats,{beats}>>
-* <<install-logstash,{ls} (optional)>>
 
-After completing the installation process, learn how to implement a system monitoring solution that uses
-{metricbeat} to collect server metrics and ship the data to {es}, search and visualize the data by using {kib}, and incorporate {ls} for additional parsing.
+TIP: {ls} adds powerful data parsing and transformation features, but usually
+isn't required. To get started with {ls}, see
+{logstash-ref}/getting-started-with-logstash.html[Getting Started with {ls}].
+
+After completing the installation process, learn how to implement a system
+monitoring solution that uses {metricbeat} to collect server metrics and ship
+the data to {es}. Then use {kib} to search and visualize the data.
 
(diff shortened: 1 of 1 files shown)


dedemorton:

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -6,10 +6,14 @@ Looking for an {stack} ("ELK") guide that shows how to quickly install and confi
 * <<install-elasticsearch,{es}>>
```
bmorelli25, line 6: On line 4:

> Looking for an {stack} ("ELK") guide that shows how to quickly install and configure the {stack}? You're in the right place! You can install the {stack} on a single VM, or even on your laptop:

Do we still want to call this an ELK guide if the L is excluded?

dedemorton: I _think_ we do. I know it seems odd, but lots of users still search on "ELK" when they're first getting started. So here's the backstory:

Originally I used "ELK tutorial" in the opening paragraph of this topic because forum users kept searching on "ELK tutorial" and finding a topic on the Digital Ocean website that recommended old Beats settings. These settings were causing major problems. Adding "ELK tutorial" pushed us up quite a bit in the rankings.

It looks like @szabosteve removed "tutorial" from the topic, but that might be OK because the Digital Ocean tutorial isn't completely broken now.

dedemorton:

docs/en/getting-started/get-started-stack.asciidoc
szabosteve: @dedemorton 
Yes, it was me back in the DEVO days. Our working group dealt with Getting started/Tutorial/Quickstart definitions and content. As part of that effort, we tried to streamline the wording of the respective pieces. Sometimes it was just to remove a word that defines the function of the piece, like here. I wasn't aware that time the reasoning behind calling it a tutorial or that you are the author of this piece, DeDe. Sorry for the inconvenience and thanks for the heads-up!

dedemorton: @szabosteve Yah that makes sense. In this case, I probably should have added a comment to the file to explain why I was doing something that was a little non-standard. :-) I think it makes sense to retain "ELK" here for now...especially since we will be replacing this GS at some point soonish.

PR #1033, opened 2020-05-05
[Docs][SIEM]Adds field reference list

benskelker:
Lists all fields used by SIEM.

[Preview](http://stack-docs_1033.docs-preview.app.elstc.co/guide/en/siem/guide/master/siem-field-reference.html)

PR #1103, opened 2020-05-26
[Docs][SIEM]Updates required fields section

benskelker:
Need to double-check that all listed fields are actually required - especially non-ECS field.

Original PR that added the fields: https://github.com/elastic/stack-docs/pull/1033

[Preview](http://stack-docs_1103.docs-preview.app.elstc.co/guide/en/siem/guide/master/siem-field-reference.html)

### DIFF
--- a/docs/en/siem/field-ref.asciidoc
+++ b/docs/en/siem/field-ref.asciidoc
@@ -2,27 +2,23 @@
 [chapter, role="xpack"]
 = SIEM field reference guide
 
-This section lists the fields that can be used to populate {siem-soln} data.
+This section lists ECS fields the {siem-app} uses to display data.
 
-IMPORTANT: Unless noted otherwise, all listed fields are {ecs-ref}[ECS fields].
-It is recommended to use {beats} to ship your data. If you cannot use {beats},
-use ECS fields whenever possible, as in future releases this list is likely to
-change and {siem-soln} is expected to adopt more ECS fields. To help reduce
-time spent on updating ingestion pipelines in the future, see
-{ecs-ref}/ecs-converting.html[how to map data to ECS].
-
-[float]
-[[siem-global-fields]]
-== Global fields
-
-Fields required for populating SIEM data:
-
-* `@timestamp`
-* `host.name`
+IMPORTANT: It is recommended to use {beats} to ship your data. When you use
+{beats}, manual field mapping is not required and all {siem-app} data is
+populated. If you cannot use {beats}, map your data to {ecs-ref}[ECS fields]
+(see {ecs-ref}/ecs-converting.html[how to map data to ECS]).
 
 [float]
 [[siem-host-fields]]
 == Host fields
+
+These fields must be mapped to display host data in the {siem-app}:
+
+* `@timestamp`
+* `host.name`
+
+These fields can be mapped to display additional host data in the {siem-app}:
 
 * `cloud.instance.id`
 * `cloud.machine.type`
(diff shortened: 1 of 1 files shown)


docs/en/siem/field-ref.asciidoc
```diff
@@ -2,15 +2,17 @@
 [chapter, role="xpack"]
 = Field reference and mappings
 
-This section lists all fields required to populate {siem-soln} data.
+This section lists the fields required to populate all {siem-soln} data.
```
MikePaquette, line 5: I still think "required" needs clarification or removal.  As an alternative, perhaps something like:  "This section lists all fields used by the current version of the SIEM app when it queries, aggregates, filters, displays, or renders event data."

docs/en/siem/field-ref.asciidoc
```diff
@@ -2,15 +2,17 @@
...
 
-NOTE: Unless noted otherwise, all listed fields are {ecs-ref}[ECS fields]. 
+NOTE: Unless noted otherwise, all listed fields are {ecs-ref}[ECS fields].
+Whenever possible, it is recommended to use ECS fields. In future releases,
+{siem-soln} will adopt more ECS fields and non-ECS fields will be deprecated. To reduce time spent on updating you ingestion pipelines in the future, see {ecs-ref}/ecs-converting.html[how to convert data to ECS].
```
MikePaquette, line 9: - suggest removing "whenever possible" to make this more clear and authoritative.
- suggest changing "will adopt" to "is expected to adopt"
- suggest removing the mention of deprecation, as we're not sure if we'll actually deprecate their use.  It is true that we hope to remove all dependencies on non-ECS fields, but that's not really the same thing.
- suggest change "updating you" to "updating your"

I think there's still a message that is not getting across.  Here's what I think it should say, in rough terms:

> Here's the list of fields required by the current version of the SIEM app. (This should have very few, if any, non-ECS fields)
> When building your ingest pipelines, do NOT limit your implementation to only these required fields, but instead populate as many ECS fields as possible/practical, since future versions of the SIEM app are expected to adopt additional ECS fields, and if you do only the required ones, you'll have to update your ingest pipelines in the future.

docs/en/siem/field-ref.asciidoc
```diff
@@ -2,15 +2,17 @@
...
 
 [float]
 [[siem-global-fields]]
 == Global fields
 
-Fields required for populating all SIEM data:
+Fields required for populating SIEM data:
```
MikePaquette, line 15: same comment as above about "required"

docs/en/siem/field-ref.asciidoc
benskelker: @MikePaquette 
@XavierM and I tried to only list fields that are relevant for ingesting data, not internal operations. What do you think about removing any fields that are not required for ingesting data? I think this section should just list all fields required to fully populate SIEM. That way, it is a complete reference guide. What do you think?

docs/en/siem/field-ref.asciidoc
benskelker: Thanks @MikePaquette - I've updated the text slightly (and fixed the typo), but I think this complicates the issue for users. After this PR is merged, only required non-ECS fields will be listed. If we start suggesting that users should create a futureproof pipeline, do we need to give them more concrete information? If their pipeline now maps 1 source field to 2 target fields (one ECS and one non-ECS), won't they want to fix that anyway?

docs/en/siem/field-ref.asciidoc
benskelker: I think these two fields are essential for getting meaningful data into SIEM.

docs/en/siem/field-ref.asciidoc
benskelker: Updated, @MikePaquette please let me know if the new text is OK with you. thanks

docs/en/siem/field-ref.asciidoc
```diff
@@ -1,23 +1,24 @@
...
 [chapter, role="xpack"]
-= Field reference and mappings
+= SIEM field reference guide
 
-This section lists all fields required to populate {siem-soln} data.
+This section lists ECS fields the {siem-app} uses to display data.
 
-NOTE: Unless noted otherwise, all listed fields are {ecs-ref}[ECS fields]. 
+IMPORTANT: It is recommended to use {beats} to ship your data. When you use
+{beats}, manual field mapping is not required and all {siem-app} data is
```
jamiehynds, line 8: I'd make it clearer that if you are using any of our out-of-the-box [Beats modules](https://www.elastic.co/guide/en/beats/filebeat/master/filebeat-modules.html), the data is already in ECS. If you are ingesting data from a custom data source, mapping to ECS is still required. Could link to [ECS Mapper](https://github.com/elastic/ecs-mapper) which simplifies the mapping process.

PR #1109, opened 2020-05-26
[DOCS] Expand an example from the Docker section of the Getting Started Guide

mbarretta:
Expands the example visible on [this page of the docs](https://www.elastic.co/guide/en/elastic-stack-get-started/current/get-started-docker.html)

Based on [this stackoverflow question](https://stackoverflow.com/questions/62010914/what-is-the-syntax-for-elasticsearch-hosts-in-docker-compose) and a conversation with its author on our public Slack channel, I thought it made sense to flesh out the example to show how to set multiple values in YAML in object format generally and for `ELASTICSEARCH_HOSTS` specifically.


jrodewig:
Thanks for the PR @mbarretta. Apologies for the delay in responding and reviewing. I've addressed these concerns in #1559.

PR #1175, opened 2020-06-15
Edit ingest management getting started guide

dedemorton:
Summary of changes:

- Apply edits to align with Elastics style guidelines for writing about UIs.
- Move all prereqs to a single section and added to the list.
- Change procedures to use numbered lists for improved scanability.
- Movedstandalone steps to appear after fleet steps as requested by @mostlyjason.

### DIFF
--- a/docs/en/ingest-management/getting-started.asciidoc
+++ b/docs/en/ingest-management/getting-started.asciidoc
@@ -51,7 +51,7 @@ setting must be `true`.
 
 * A {kib} configuration with the following capabilities enabled:
 +
-- {kibana-ref}/configuring-security.html[{kib} security]. The `xpack.security.enabled`
+- {kibana-ref}/using-kibana-with-security.html[{kib} security]. The `xpack.security.enabled`
 setting must be `true`.
 - {kibana-ref}/configuring-tls.html[TLS] or the
 `pack.ingestManager.fleet.tlsCheckDisabled` setting must be `true`.
--- a/docs/en/ingest-management/getting-started.asciidoc
+++ b/docs/en/ingest-management/getting-started.asciidoc
@@ -1,94 +1,145 @@
 [[ingest-management-getting-started]]
 [role="xpack"]
-= Getting started
+= Get started with ingest management
 
-//TODO: Technical edit.
-
-//TODO: Add a summary.
+++++
+<titleabbrev>Get started</titleabbrev>
+++++
 
 experimental[]
 
-To use {ingest-management}, you need an {es} cluster and {kib} (version 7.8 for the experimental release)
-with a basic license. See {stack-gs}/get-started-elastic-stack.html[Getting started with the {stack}].
+This guide describes how to get started with the new {ingest-management}
+capabilities available in this release.
 
-//TODO: Add cloud info
+// tag::experimental-warning[]
+This early release of {ingest-management} is experimental. We invite you to
+install and test these capabilities **in a test environment**. You might run
+into problems and need to modify your setup to get this feature running. Please
+do not enable or use {ingest-management} in a production environment.
+// end::experimental-warning[]
 
-IMPORTANT: Since this is an experimental release, it should only be used in a dedicated test environment
-that can be deleted when you are done. Please do not enable or use {ingest-management} in a production
-environment.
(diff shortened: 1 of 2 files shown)


docs/en/ingest-management/getting-started.asciidoc
```diff
@@ -1,159 +1,191 @@
...
+This early release of {ingest-management} is experimental. We invite you to
+install and test these capabilities **in a test environment**. You might run
+into problems and need to modify your setup to get this feature running. Please
+do not enable or use {ingest-management} in a production environment.
+// end::experimental-warning[]
+
+//REVIEWERS: We have several different statements like this one that say basically the same
+//thing but use slightly different words. Let's create a standard description that I
+//can include when we need to add this statement. Let me know if what I have here ^^
+//is good.
```
skh, line 25: :+1: from me.

docs/en/ingest-management/getting-started.asciidoc
```diff
@@ -1,159 +1,191 @@
...
++
+- {kibana-ref}/using-kibana-with-security.html[{kib} security]. The `xpack.security.enabled`
+setting must be `true`.
+- {kibana-ref}/configuring-tls.html[TLS] or the
+`pack.ingestManager.fleet.tlsCheckDisabled` setting must be `true`.
 
-For feedback and questions, please contact us https://ela.st/ingest-manager-feedback[here]. 
+//REVIEWERS: I've added the Kibana requirements because they weren't stated. Let
+//me know if you'd rather that I leave those out. I personally have concerns
+//about telling users to disable TLS checks.
```
skh, line 61: Cc @nchaulet about the TLS checks and when exactly they need to be disabled. Maybe adding more reasoning why this is necessary in some situations helps? (If not, please ignore.)

docs/en/ingest-management/getting-started.asciidoc
```diff
@@ -1,159 +1,191 @@
...
-In this Guide we assume that you have `nginx` running on some of your infrastructure, and want to
-collect logs and metrics from it. To do so, first install the `nginx` integration provided by Elastic.
+{ingest-manager} provides _integrations_ that bundle various assets needed to
+ingest and visualize data.
 
-In the {ingest-manager} app, navigate to **Integrations** and use the search bar to find the Nginx integration.
+In this guide, we assume that you have `nginx` running on some of your
+infrastructure, and want to collect logs and metrics from it. To do so:
 
+. Install the `nginx` integration provided by Elastic:
```
skh, line 107: When I wrote this, I had an extra step about installing the integration (can be done from the 'settings' section in the integration details page) and only then explained how to create the data source.

As creating the data source installs the integration in the background this step wasn't necessary and I removed it. Maybe this line should be rephrased or removed so that users don't look for an explicit installation step they don't need.

docs/en/ingest-management/getting-started.asciidoc
```diff
@@ -162,30 +194,109 @@ The {agent} is currently in Experimental and should not be used in production
...
+tab, where you should see the newly enrolled agent.
++
 [role="screenshot"]
 image::images/kibana-ingest-manager-fleet-agents.png[{ingest-manager} app showing enrolled agents]
 
-When you want to unenroll an agent, choose **Unenroll** from the **Actions** menu for this agent. This will invalidate the API key the agent is using to connect to {es}. The {agent} will continue to run, but will not be able to send data, and show this error instead:
+//REVIEWERS: This is super cool, but we kind of gloss over the notion that this
+//only works if the logs are stored in the location expected for the service
+//your monitoring. Do we want to tell users how to override default paths, or
+//is that unnecessary for this release?
```
skh, line 213: Good point. I didn't leave it out on purpose, I think it would be a useful addition.

docs/en/ingest-management/getting-started.asciidoc
dedemorton: Thanks! If necessary, I can open a follow-up PR after this PR is merged.

nchaulet: The scenario where the user wants to disable the TLS check, is if Kibana is behind a proxy that terminate the ssl connection.

PR #1180, opened 2020-06-16
Fix getting started es port mappings

marcofranssen:
The es PORT in the container is 9200


jrodewig:
Thanks for raising this @marcofranssen. Apologies for the delay in reviewing and responding.

We've updated the example file in #1559, which should address these concerns.

Thanks again!

PR #1183, opened 2020-06-16
[7.x] Edit ingest management overview (#1168)

dedemorton:
Backports the following commits to 7.x:
 - Edit ingest management overview (#1168)


dedemorton:
Closing without merging because the previous backport is missing.

PR #1192, opened 2020-06-16
variables in shell formatting are not expanded

EricDavisX:
hi - @dedemorton is this a normal thing or something new-ish?  In the 'shell' formatting usage the {agent} variable is not expanded to 'Elastic Agent' so I did this up.  

I don't know the ascii-docs format hardly at ll, if its not the right way to fix, thats ok - we can fix it better, just seemed easy to propose this in case it is right.  :)
 - should it get merged to master first and then to 7.8?  that's the usual process, I know...  update this if appropriate.  or i can, thanks much.

screenshot:
Screen Shot 2020-06-16 at 6 51 17 PM

### DIFF
--- a/docs/en/ingest-management/getting-started.asciidoc
+++ b/docs/en/ingest-management/getting-started.asciidoc
@@ -183,7 +183,7 @@ that directory.
 ----
 ./elastic-agent enroll KIBANA_URL ENROLLMENT_KEY
 
-The {agent} is currently in Experimental and should not be used in production
+The Elastic Agent is currently in Experimental and should not be used in production
 This will replace your current settings. Do you want to continue? [Y/n]:
 ----
 


dedemorton:
Nice catch! This is a bug. The `agent` attribute is valid, but attribute resolution needs to be turned on for the code block. I'll add this to my list.

Closing this PR because it's against the wrong branch (we want to make all code changes in master then backport to 7.8 and 7.x) and the fix isn't quite right. 

If Brandon M doesn't notice this when he tests the docs (I bet he will), I'll fix this tomorrow when I do my cleanup. Thanks!

dedemorton:
Note that this problem is fixed in this commit: https://github.com/elastic/stack-docs/pull/1194/commits/49f49328431428b45200921e83fd3d4074d26426. Thanks again for finding this!

EricDavisX:
right on!  glad it is fixed up. :)

PR #1214, opened 2020-06-24
Adds alphabetical links to Glossary

lcawl:
Depends on https://github.com/elastic/kibana/pull/69721 and https://github.com/elastic/docs/pull/1875

This PR adds alphabetical sections and links to the Glossary (https://www.elastic.co/guide/en/elastic-stack-glossary/current/index.html). It also moves the "Introduction" page into the "Terminology" page, so that there's only one page in this book.

### Preview

https://stack-docs_1214.docs-preview.app.elstc.co/guide/en/elastic-stack-glossary/master/terms.html


lcawl:
> The A, B, C headings are showing up in the right side bar.

Yes, unfortunately discrete attributes don't seem to affect H3 titles in that "On this page" section after all.

> Also, how about removing the heading "Introduction"? I'm not sure it's needed.

I kept it in case there were links to that existing page.  I can try removing it and see if anything breaks, however.

lcawl:
I will revert to sections instead of description lists since I don't like the indentation issues we're hitting per https://github.com/elastic/docs/issues/1876

gtback:
> I will revert to sections instead of description lists

I feel like using section headers for each letter would be more semantically correct than description links anyway. 👍 

I imagine there's a bug in how Asciidoctor tries to infer the list hierarchy, because it seems like it would be pretty complex.

PR #1210, opened 2020-06-23
[DOCS] Split fields and metrics content in Metrics monitoring guide

EamonnTP:
This PR separates the fields and metrics content in the Metircs monitoring guide and lists the fields required for populating the Metrics UI in Kibana. 

### Preview

Fields: https://stack-docs_1210.docs-preview.app.elstc.co/guide/en/metrics/guide/master/metrics-fields-reference.html

Metrics: https://stack-docs_1210.docs-preview.app.elstc.co/guide/en/metrics/guide/master/infrastructure-metrics.html

### Related issue

https://github.com/elastic/observability-dev/issues/223

PR #1203, opened 2020-06-18
[DOCS] Add ECS fields to Logs monitoring guide

EamonnTP:
This PR adds the ECS fields required for populating the Logs UI in Kibana. The PR also addresses minor issues in the Logs installation instructions that were called out in this [issue](https://github.com/elastic/stack-docs/issues/646).

### Preview 

Installation: https://stack-docs_1203.docs-preview.app.elstc.co/guide/en/logs/guide/master/install-logs-monitoring.html

ECS fields: https://stack-docs_1203.docs-preview.app.elstc.co/guide/en/logs/guide/master/logs-fields-reference.html

### Related issues

Closes https://github.com/elastic/stack-docs/issues/646
https://github.com/elastic/observability-dev/issues/223

PR #1216, opened 2020-06-25
[DOCS] Add placeholders for Observability docs

EamonnTP:
This PR adds content placeholders for the forthcoming Observability docs. This directory and the files will be included in the migration of the Logs and Metrics directories to the new Observability repo. 

The following PRs need to be merged before migrating the Logs and Metrics directories:

https://github.com/elastic/stack-docs/pull/1210
https://github.com/elastic/stack-docs/pull/1203

### Related PRs:

https://github.com/elastic/docs/pull/1877

### DIFF
--- a/docs/en/observability/observability-introduction.asciidoc
+++ b/docs/en/observability/observability-introduction.asciidoc
@@ -1,0 +1,9 @@
+[[observability-introduction]]
+[role="xpack"]
+
+= Observability 
+++++
+<titleabbrev>What is Elastic Observability?</titleabbrev>
+++++
+
+//TODO: Add content
--- a/docs/en/observability/observability-introduction.asciidoc
+++ b/docs/en/observability/observability-introduction.asciidoc
@@ -1,9 +1,36 @@
 [[observability-introduction]]
 [role="xpack"]
-
-= Observability 
-++++
-<titleabbrev>What is Elastic Observability?</titleabbrev>
-++++
+== What is Elastic Observability? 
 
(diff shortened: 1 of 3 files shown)


docs/en/observability/observability-introduction.asciidoc
```diff
@@ -1,9 +1,36 @@
 [[observability-introduction]]
 [role="xpack"]
-
-= Observability 
-++++
-<titleabbrev>What is Elastic Observability?</titleabbrev>
-++++
+== What is Elastic Observability?
```
dedemorton, line 3: I know this is all placeholder stuff, but we should start thinking about how we want to structure the content.

1 - We definitely want a one to one mapping between topics and HTML output. So one asciidoc file = one HTML file.

2 - To control TOC nesting, we should explore using level offsets. I've done this in the ingest management docs with mixed success. (The strategy worked for +1 and +2 offsets, but broke for anything nested more deeply.) We should follow up with other teams to see if they've had more success, but I think it's the best way forward to make it easier to decouple TOC/navigation from heading levels.

bmorelli25:
Small note: I'd also like to see us remove the unnecessary `/en/` directory in our docs

dedemorton:
Let's bring up the question about `/docs/en/` with other writers in the org. Eventually we will need language-specific folders for translation.

bmorelli25:
Ahh, I didn't realize that was still something we were planning on doing.

PR #1219, opened 2020-06-26
remove unnecessary options for elasticsearch-setup-passwords

notok:
Removing `xpack.security.http.ssl.*` options because the following error occurs when running `elasticsearch-setup-passwords` with these.

```
ERROR: setting [xpack.security.http.ssl.certificate] already set, saw [certificates/es01/es01.crt] and [/usr/share/elasticsearch/config/certificates/es01/es01.crt]
```


lcawl:
Thanks for fixing this!  It's related to https://github.com/elastic/elasticsearch/pull/50221 so I'll forward-fit to 7.x and 8.0 and backport to 7.7 and 7.6 branches.

PR #1230, opened 2020-06-30
[Docs][SIEM]Adds ml rule prerequisites 

benskelker:
Adds ml rule creation requirements to the Detections prerequisites as users are having trouble finding this info.

[Preview](https://stack-docs_1230.docs-preview.app.elstc.co/guide/en/siem/guide/master/detection-engine-overview.html#detections-permissions)

### DIFF
--- a/docs/en/siem/detections/detection-engine-intro.asciidoc
+++ b/docs/en/siem/detections/detection-engine-intro.asciidoc
@@ -162,7 +162,7 @@ For *all* deployments (on-premises and hosted):
 ** Write permissions for the `.siem-signals-<space name>` index, such as 
 `create` `create_doc`, `write`, `index`, and `all`
 (see {ref}/security-privileges.html#privileges-list-indices[Indices privileges]).
-** To run and create {ml} rules, you must have the
+** To create {ml} rules, you must have the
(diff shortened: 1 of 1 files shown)


docs/en/siem/detections/detection-engine-intro.asciidoc
```diff
@@ -162,6 +162,9 @@ For *all* deployments (on-premises and hosted):
 ** Write permissions for the `.siem-signals-<space name>` index, such as 
 `create` `create_doc`, `write`, `index`, and `all`
 (see {ref}/security-privileges.html#privileges-list-indices[Indices privileges]).
+** To create {ml} rules, you must have the
```
rylnd, line 165: While this is listed under "To create and modify detection rules..." I worry that it might be confusing to call out creation specifically here. I would recommend changing it to either:

```suggestion
** To create or modify {ml} rules, you must have the
```
or
```suggestion
** For {ml} rules, you must have the
```

benskelker: The outcome is...

** To create or modify {ml} rules, you must have the

Thanks @rylnd

PR #1271, opened 2020-07-13
[DOCS] Add model memory tip for ML modules

lcawl:
This PR adds a tip in https://www.elastic.co/guide/en/machine-learning/master/ootb-ml-jobs.html for dealing with model memory limits.

### Preview

* https://stack-docs_1271.docs-preview.app.elstc.co/guide/en/machine-learning/master/anomaly-detection-scale.html#set-model-memory-limit
* https://stack-docs_1271.docs-preview.app.elstc.co/guide/en/machine-learning/master/ootb-ml-jobs.html

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
@@ -165,7 +165,7 @@ is based on the analysis configuration details for the job and cardinality
 estimates, which are derived by running aggregations on the source indices as
 they exist at that specific point in time. 
 
-If you change the resources available on your {ml} nodes or the make significant 
+If you change the resources available on your {ml} nodes or make significant 
 changes to the characteristics or cardinality of your data, the model memory 
 requirements might also change. You can update the model memory limit for a job 
 while it is closed. If you want to decrease the limit below the current model 
--- a/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
@@ -180,8 +180,8 @@ each job. If you reach the hard limit and are concerned about the missing data,
 ensure that you have adequate resources then clone and re-run the job with a 
 larger model memory limit. For more information about retrieving the memory 
 status, see the
-{ref}ml-get-job-stats.html[get {anomaly-jobs} stats API] or the
-{ref}ml-get-snapshot.html[get model snapshots API].
+{ref}/ml-get-job-stats.html[get {anomaly-jobs} stats API] or the
+{ref}/ml-get-snapshot.html[get model snapshots API].
 
 [discrete]
 [[pre-aggregate-data]]
--- a/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
@@ -162,14 +162,21 @@ The `model_memory_limit` job configuration option sets the approximate maximum
 amount of memory resources required for analytical processing. When you create
 an {anomaly-job} in {kib}, it provides an estimate for this limit. The estimate 
 is based on the analysis configuration details for the job and cardinality 
-estimates for the fields it references. You can also obtain this information
-from the
-{ref}/ml-estimate-model-memory.html[estimate {anomaly-jobs} model memory API].  
+estimates, which are derived by running aggregations on the source indices as
+they exist at that specific point in time. 
 
-If you change the resources available on your {ml} nodes or the characteristics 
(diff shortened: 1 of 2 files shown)


sophiec20:
We set a mem limit for each job so that there is finer grain control to enable smoother operations for multiple jobs running on the same node and so we can make better decisions wrt job placement in multi-node environments.

We set the memory limit (since 7.8) based on an analysis of the data characteristic in the source indices at the point we create the job. This may not end up being ideal in cases where the ML jobs were created before ingest firehoses were turned on or ramped up. Or if the ML jobs were created on a low-spec node, and ML nodes have subsequently been given more resources. Or if the data characteristics change e.g. a large increase geo location of the user base would change the cardinality of location based analysis and the mem requirements for those jobs will increase as new data is seen.

Users should proactively increase the estimated model mem size if the above is true. Or they may want to because they know their env is "large" and they have resources to spare.

Otherwise, during the operations of the job, a mem status warning may occur. If `SOFT_LIMIT` we would advise to edit the job and increase the memory limit. With `HARD_LIMIT`, it depends. If you care about the entities that have not been modelled, then you should clone and restart. If you are happy to continue, then you can increase the model limit.

This is applicable for ootb jobs as well as other jobs. I think it is good to highlight on this page for ootb users and I think a topic on memory management would be useful in core ML docs - although this does not need to be the same PR.

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -159,18 +159,29 @@ start the {dfeed} for the change to be applied.
...
-manually. The estimation is based on the analysis configuration details for the 
-job and cardinality estimates for the fields it references. You can update the 
-memory settings of an existing job, but the job must be closed.
-
+amount of memory resources required for analytical processing. When you create
+an {anomaly-job} in {kib}, it provides an estimate for this limit. The estimate 
+is based on the analysis configuration details for the job and cardinality 
+estimates for the fields it references. You can also obtain this information
+from the
+{ref}/ml-estimate-model-memory.html[estimate {anomaly-jobs} model memory API].
```
droberts195, line 167: It's worth adding that the cardinality estimates used by Kibana come from running aggregations against the source indices of the datafeed at the time Kibana is being used to create the job.  If these are not reflective of the cardinality the job will experience while running (for example because ingest hasn't yet started at the time the job is created) then the memory estimate will be wrong.

If you use the estimate anomaly detection jobs model memory API then you must provide your own cardinality estimates.

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -159,18 +159,29 @@ start the {dfeed} for the change to be applied.
...
+an {anomaly-job} in {kib}, it provides an estimate for this limit. The estimate 
+is based on the analysis configuration details for the job and cardinality 
+estimates for the fields it references. You can also obtain this information
+from the
+{ref}/ml-estimate-model-memory.html[estimate {anomaly-jobs} model memory API].  
+
+If you change the resources available on your {ml} nodes or the characteristics 
+of your data, the memory requirements might also change. You can increase the 
+model memory limit for an existing job while it is closed. To decrease the limit, 
+you must clone and re-run the job.
```
droberts195, line 172: You can also decrease the model memory limit while the job is closed, providing you do not reduce it below the size reported in the latest model size stats.  (The ability to reduce model memory limit was added in 6.3.)

lcawl: Thanks!  I've tried to make that clearer in the API too via https://github.com/elastic/elasticsearch/pull/74517

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
lcawl: Thanks, I've interwoven those details in these paragraphs.

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -159,18 +159,33 @@ start the {dfeed} for the change to be applied.
...
-job and cardinality estimates for the fields it references. You can update the 
-memory settings of an existing job, but the job must be closed.
-
+amount of memory resources required for analytical processing. When you create
+an {anomaly-job} in {kib}, it provides an estimate for this limit. The estimate 
+is based on the analysis configuration details for the job and cardinality 
+estimates, which are derived by running aggregations on the source indices as
+they exist at that specific point in time. 
+
+If you change the resources available on your {ml} nodes or the make significant
```
lcawl, line 168:
```suggestion
If you change the resources available on your {ml} nodes or make significant 
```

droberts195 (approved):
LGTM

PR #1311, opened 2020-07-29
Activate include for Beats highlights

dedemorton:


dedemorton:
No actual content here and the build passes, so I'm merging.

PR #1349, opened 2020-08-31
docs: add Observability terms to glossary

bmorelli25:
This PR adds the following Observability terms to the Elastic Glossary:

* **Observability**
* **Agent (APM)**
* **Server (APM)**
* **Distributed tracing**
* **Instrumentation**
* **Real User Monitoring (RUM)**
* **Span**
* **Trace**
* **Transaction**
* **Ingestion**
* **Integration**
* **Module**
* **Monitor**
* **Uptime**

I didn't source these terms from the `observability-docs` repo because we don't have any short-term plans to host an Observability glossary.

Closes https://github.com/elastic/observability-docs/issues/34.

PR #1351, opened 2020-09-03
[7.x] docs: add Observability terms to glossary (#1349)

bmorelli25:
Backports the following commits to 7.x:
 - docs: add Observability terms to glossary (#1349)


bmorelli25:
It looks like the glossary only exists in `master`.

PR #1359, opened 2020-09-10
[DOCS] Add feature importance to classification example

lcawl:
Related to https://github.com/elastic/kibana/pull/73561

This PR drafts changes to the classification example such that it includes feature importance explanations.

It will be backported to 7.10 and does not take into consideration changes in 7.11 and later for https://github.com/elastic/kibana/issues/77874

### Preview

https://stack-docs_1359.docs-preview.app.elstc.co/guide/en/machine-learning/master/flightdata-classification.html


docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -343,50 +350,95 @@ The snippet below shows a part of a document with the annotated results:
...
-data point to the class with the highest score maximizes the minimum recall of 
-any class.
+//Does this mean the sum of the feature importance values for false in this
+example should equal the logit(p), where p is the class_probability for false?
+//Does this imply that the feature importance value itself is the result of a logit function? Or that we use the function to merely represent the distribution of feature importance values?
 ////
-====
+While the probability of a class ranges between 0 and 1, its log-odds range 
+between negative and positive infinity. In {kib}, the decision path for each
+class starts near zero, which represents a class probability of 0.5.
```
valeriy42, line 432: This is unfortunately a bit more complicated: 0 would represent the class probability of a constant baseline. It relates to the average class probability for the selected class (in Kibana UI) over entire training set.

If you select `Canceled` as the target variable in flight data, this nuance becomes obvious. Since there are many more data points with `Canceled = False`, let's assume that the average class probability over the entire training set would be something like `0.92`. This means that if the class probability of a data point is larger than `0.92` (for example `0.98`) than the decision path will go to the right (sum of feature importances is positive). On the other hand, if the class probability is smaller than `0.92` (for example 0.84), than the decision path will go to the left (sum of feature importances is negative).

docs/en/stack/ml/df-analytics/flightdata-classification.asciidoc
```diff
@@ -343,50 +350,95 @@ The snippet below shows a part of a document with the annotated results:
...
+// Is this true for multi-class classification or just binary classification?
```
valeriy42, line 433: Yes, it's true for both multi-class and binary since in the UI you are selecting a class of interest from a drop-down menu.

PR #1360, opened 2020-09-11
[DOCS] Adds feature importance regression example

lcawl:
Related to elastic/kibana#73561

This PR drafts changes to the [regression example](https://www.elastic.co/guide/en/machine-learning/master/flightdata-regression.html) such that it includes feature importance explanations.

In particular, it suggests choosing a feature importance value in step 2g then shows an example of a decision plot in the "Viewing regression results" section.

It also updates the destination index names (since by default the index name=job name when you use Kibana) and adds a missing query in the wizard screenshot (which previously only existed in the API version of the example).

### Preview

https://stack-docs_1360.docs-preview.app.elstc.co/guide/en/machine-learning/master/flightdata-regression.html


valeriy42 (approved):
LGTM.

szabosteve (approved):
LGTM, thanks!

PR #1404, opened 2020-10-12
Update flightdata-classification.asciidoc

mrklaney:
This paragraph is explaining the bottom row of a confusion matrix.  It correctly states the first column as 'false negative,' but then incorrectly says the second column (i.e. bottom right) is a true positive. TP is in the upper left.  See https://en.wikipedia.org/wiki/Confusion_matrix    It's not called a 'confusion' matrix for nothing!   Easy to mislabel. This is particularly confusing (haha) because the categories are True and False, which muddles the idea of Positive (row 1) vs Negative (row 2) predictions.


valeriy42:
The confusion matrix in the Wikipedia is the transposed version of our confusion matrix: we have predicted labels as columns, wikipedia has predicted labels as rows. 

AFAIK, the position of "true positive" in the first row  and "true negative" is the second row of the matrix is not dogmatic, see for example [the wikipedia article on type 1 and type 2 errors](https://en.wikipedia.org/wiki/Type_I_and_type_II_errors#Type_I_error).

PR #1439, opened 2020-11-06
[DOCS] Adds Moving trained models between clusters to docs

szabosteve:
## Overview

This PR adds a subsection about how to move trained models between clusters to the `Trained models` section.

### Preview

[Trained models – Moving a trained model between clusters](https://stack-docs_1439.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-trained-models.html#move-between-clusters)

### DIFF
--- a/docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
+++ b/docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
@@ -48,7 +48,7 @@ trained models.
 
 [source,console]
 --------------------------------------------------
-GET _ml/inference/
+GET _ml/trained_models/
 --------------------------------------------------
 // TEST[skip:setup kibana sample data]
 
@@ -57,9 +57,9 @@ The API response contains the `model_id` of the trained models. Check the
 call in the next step.
 --
 
-2. Use the {ref}/get-inference.html[GET trained model API] to get the trained 
-model definition. You need to specify the following query parameters in the 
-call:
+2. Use the {ref}/get-trained-models.html[GET trained model API] to get the 
+trained model definition. You need to specify the following query parameters in 
+the call:
 +
 --
 * `for_export`: This parameter allows the model to be in an acceptable format to 
@@ -79,7 +79,7 @@ The following call is an example to get the trained model definition. (Replace
 
 [source,console]
 --------------------------------------------------
-GET _ml/inference/<model_id>?for_export=true&include_model_definition=true&decompress_definition=false
+GET _ml/trained_models/<model_id>?for_export=true&include_model_definition=true&decompress_definition=false
 --------------------------------------------------
 // TEST[skip:setup kibana sample data]
 
@@ -90,8 +90,8 @@ information.
 
 3. Copy the content of `trained_model_configs`.
 
-4. Use the {ref}/put-inference.html[Create trained model API] in the cluster you 
-want to move the trained model to. Paste the content of the 
+4. Use the {ref}/put-trained-models.html[Create trained model API] in the 
+cluster you want to move the trained model to. Paste the content of the 
 `trained_model_configs` to the request body of the API call.
 +
 --
@@ -103,7 +103,7 @@ The following call is an example to create the trained model. The actual
 ====
 [source,console]
 --------------------------------------------------
-PUT _ml/inference/<my_model_id>
+PUT _ml/trained_models/<my_model_id>
 {
    "compressed_definition":"<definition value of the trained model>",
    "tags":[
--- a/docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
+++ b/docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
@@ -216,7 +216,7 @@ examples below.
 
 [discrete]
 [[move-trained-model-to-es]]
-=== Moving a model to {es}
+=== Moving a model to the {stack}
 
 It is possible to add a model to your {es} cluster even if the model is not 
 trained by {dfanalytics}.
--- a/docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
+++ b/docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
@@ -33,12 +33,11 @@ Alternatively, you can use APIs like
 [[move-between-clusters]]
 == Moving a trained model between clusters
 
-It is a common scenario that the {ml} models are trained in a development
-or test environment and then used in a production environment. 
-In this case, you need to move your trained 
-model from one cluster to another. The trained model APIs enable you to move 
-your trained model between clusters. The following description shows you the 
-process step by step.
+It is a common scenario that the {ml} models are trained in a development or 
+test environment and then used in a production environment. In this case, you 
+need to move your trained model from one cluster to another. The trained model 
+APIs enable you to move your trained model between clusters. The following 
+description shows you the process step by step.
 
 1. (Optional) In the cluster where you trained the model, make the call below by 
 using the console in **Dev Tools** to get the configuration information of your 
@@ -65,7 +64,8 @@ the call:
 * `for_export`: This parameter allows the model to be in an acceptable format to 
 be retrieved and then added to another cluster. Set it to `true`.
 
-* `include`: Set this to `definition` for the API to include the definition in the response.
+* `include`: Set this to `definition` for the API to include the definition in 
+the response.
 
 * `decompress_definition`: It specifies in what format the included model 
 definition should be returned. Set it to `false` for getting a custom compressed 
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,223 @@ information about this process, see <<ml-supervised-workflow>> and
...
+
+1. (Optional) In the cluster where you trained the model, make the call below by 
+using the console in **Dev Tools** to get the configuration information of your 
+trained models.
++
+--
+
+[source,console]
+--------------------------------------------------
+GET _ml/inference/
```
lcawl, line 51:
```suggestion
GET _ml/inference/
```
I think this is deprecated and we should use the trained models API instead: https://www.elastic.co/guide/en/elasticsearch/reference/master/get-trained-models.html

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,223 @@ information about this process, see <<ml-supervised-workflow>> and
...
+
+4. Use the {ref}/put-inference.html[Create trained model API] in the cluster you 
+want to move the trained model to. Paste the content of the 
+`trained_model_configs` to the request body of the API call.
++
+--
+The following call is an example to create the trained model. The actual 
+`compressed_definition` value is replaced by a place holder.
+
+.Create trained model API example
```
lcawl, line 101: Looking at this example, it made me wonder which of these options were required to be the same as in the original cluster (other than compressed_definition).  Do they all have to be identical?  If not, I'm not sure if there's a lot of value to having this example and wonder if it should be removed, since the only option of importance is not shown fully anyway.

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,223 @@ information about this process, see <<ml-supervised-workflow>> and
...
+computer clipboard to copy and paste it. It is recommended to do it 
+programmatically, for example via a bash script or via Client code. You can find 
+examples below.
+--
+
+// Bash and Python examples
+
+[discrete]
+[[move-trained-model-to-es]]
+=== Moving a model to {es}
```
lcawl, line 219: Is there reason this is specifically mentioning Elasticsearch? It seems like the section is about moving it into Elastic or the Stack in general:
```suggestion
=== Moving a model to the {stack}
```

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,223 @@ information about this process, see <<ml-supervised-workflow>> and
...
+
+It is possible to add a model to your {es} cluster even if the model is not 
+trained by {dfanalytics}.
```
lcawl, line 222: If I'm understanding correctly and this is about moving a model generated outside Elastic into Elastic, I think it might be clearer by adding this qualification:
```suggestion
trained by Elastic {dfanalytics}.
```

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
szabosteve: @lcawl AFAIK, you can change the metadata in the PUT body, but it does not affect your model. However, the model configuration data will be inaccurate in this case as the model hasn't been trained with that configuration. The compressed_definition is a really long (in this example it's almost 50.000 character-long) series of characters and it is not human-readable. This is why I would omit it. I'll ask Ben about his opinion when I open the PR for review.

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,270 @@ information about this process, see <<ml-supervised-workflow>> and
...
 {ref}/delete-trained-models.html[delete trained models].
+
+
+[discrete]
+[[move-between-clusters]]
+== Moving a trained model between clusters
+
+It is a common scenario that the {ml} models are trained in a cluster that is 
+used for development and test purposes, and they are used in another cluster 
+which is the production environment. In this case, you need to move your trained
```
benwtrent, line 38:
```suggestion
It is a common scenario that the {ml} models are trained in a development
or test environment and then used in a production environment. 
In this case, you need to move your trained 
```

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,270 @@ information about this process, see <<ml-supervised-workflow>> and
...
+2. Use the {ref}/get-trained-models.html[GET trained model API] to get the 
+trained model definition. You need to specify the following query parameters in 
+the call:
++
+--
+* `for_export`: This parameter allows the model to be in an acceptable format to 
+be retrieved and then added to another cluster. Set it to `true`.
+
+* `include_model_definition`: It specifies whether the model definition is 
+returned in the response. Set it to `true`.
```
benwtrent, line 69:
```suggestion
* `include`: Set this to `definition` for the API to include the definition in the response.
```
`include_model_definition` is deprecated.

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,270 @@ information about this process, see <<ml-supervised-workflow>> and
...
+format. It is also valid to use the JSON format, but it is not optimal. As the 
+decompressed definition may be significantly larger, it is recommended to use 
+the compressed format.
+   
+The following call is an example to get the trained model definition. (Replace 
+`<model_id>` with the actual ID of the trained model.)
+
+[source,console]
+--------------------------------------------------
+GET _ml/trained_models/<model_id>?for_export=true&include_model_definition=true&decompress_definition=false
```
benwtrent, line 82:
```suggestion
GET _ml/trained_models/<model_id>?for_export=true&include=definition&decompress_definition=false
```

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
benwtrent: I agree with not fully expanding the `compressed_definition`, but it might be useful to elide the rest of the config for brevity.

szabosteve: I removed the API example in 106e319.

benwtrent (approved):

Winterflower:
@szabosteve everything looks good, but I will rerun the Python . snippets tomorrow on 7.10 to make sure everything still works and will comment back if . changes are needed

Winterflower:
Hi @szabosteve ! Just spoke to the backend team - seems there have been a few changes to how model exporting works, so I'll have to rewrite some parts of the Python. Will check-in with you tomorrow!

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,163 @@ information about this process, see <<ml-supervised-workflow>> and
...
+computer clipboard to copy and paste it. It is recommended to do it 
+programmatically, for example via a bash script or via Client code. You can find 
+examples below.
+--
+
+The following Python snippet exports the trained model that you reference to a 
+JSON file:
+
+[source, py]
+--------------------------------------------------
```
Winterflower, line 112: Hi @szabosteve ! Can we add these two lines to the snippet
```
import json

from elasticsearch import Elasticsearch
from elasticsearch.client.ml import MlClient
```

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,163 @@ information about this process, see <<ml-supervised-workflow>> and
...
+    handle.write(json.dumps(compressed_df))
+--------------------------------------------------
+// NOTCONSOLE
+
+
+The following Python snippet imports the model that stored in the JSON file to 
+a cluster:
+
+[source, py]
+--------------------------------------------------
```
Winterflower, line 127: Can we add these two lines to the top
```
import json

from elasticsearch import Elasticsearch
from elasticsearch.client.ml import MlClient
```

docs/en/stack/ml/df-analytics/ml-trained-models.asciidoc
```diff
@@ -13,16 +13,169 @@ information about this process, see <<ml-supervised-workflow>> and
...
+`model_id` of the trained model you want to move, you need to add it to the API 
+call in the next step.
+--
+
+2. Use the {ref}/get-trained-models.html[GET trained model API] to get the 
+trained model definition. You need to specify the following query parameters in 
+the call:
++
+--
+* `for_export`: This parameter allows the model to be in an acceptable format to
```
Winterflower, line 64: @davidkyle wasn't this param only added in 7.10?

PR #1449, opened 2020-11-16
[DOCS] Adds chunking_config recommendation to AD at scale

szabosteve:
## Overview

This PR expands the `Pre-aggregate your data` section in the Anomaly detection at scale piece with recommendation around `chunking_config`.

### Preview

[ADAS – Pre-aggregate your data](https://stack-docs_1449.docs-preview.app.elstc.co/guide/en/machine-learning/master/anomaly-detection-scale.html#pre-aggregate-data)

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
@@ -170,6 +170,11 @@ The statistics can be calculated in the job itself or via aggregations. It is
 more efficient to use an aggregation when it’s possible, as in this case, the 
 data node does the heavy-lifting instead of the {ml} node.
 
+You may want to use `chunking_config` to tune your search speed when your 
+{dfees} use aggregations. In these cases, set `chunking_config`.`mode` to 
+`manual` and experiment with the `time_span` value. Its maximum value is the 
+value of `search.max_buckets` which is 10,000 by default.
+
 In certain cases, you cannot do aggregations to increase performance. For 
 example, categorization jobs use the full log message to detect anomalies, so 
 this data cannot be aggregated. If you have many influencer fields, it may not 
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -170,6 +170,11 @@ The statistics can be calculated in the job itself or via aggregations. It is
 more efficient to use an aggregation when it’s possible, as in this case, the 
 data node does the heavy-lifting instead of the {ml} node.
 
+You may want to use `chunking_config` to tune your search speed when your 
+{dfees} use aggregations. In these cases, set `chunking_config`.`mode` to 
+`manual` and experiment with the `time_span` value. Its maximum value is the 
+value of `search.max_buckets` which is 10,000 by default.
```
dimitris-athanasiou, line 176: > Its maximum value is the value of `search.max_buckets` which is 10,000 by default.

I think we need to rephrase this a bit. The value of `search.max_buckets` is the number of buckets allowed in a search response. The `time_span` is a time value. So, in the case of single metric, the maximum value for `time_span` is 10,000 times the date histogram interval.

Perhaps the what we need to explain here is something along the lines of the following:

```
... and experiment with the `time_span` value. Increasing it may speed things up. However, the higher the chunking time_span, the higher number of buckets will be included in the search response. Thus, if you are hitting the `search.max_buckets` limit you can decrease the `time_span` to reduce the number of buckets per response.
```

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -170,6 +170,11 @@ The statistics can be calculated in the job itself or via aggregations. It is
 more efficient to use an aggregation when it’s possible, as in this case, the 
 data node does the heavy-lifting instead of the {ml} node.
 
+You may want to use `chunking_config` to tune your search speed when your 
+{dfees} use aggregations. In these cases, set `chunking_config`.`mode` to
```
dimitris-athanasiou, line 174: Should it be "`chunking_config`.`mode`" or "`chunking_config.mode`? I see we include the dot in the quotes below for the `search.max_buckets` limit.

szabosteve:
@dimitris-athanasiou Thanks for your suggestions. I updated the PR accordingly.

dimitris-athanasiou (approved):
LGTM

PR #1473, opened 2020-11-25
[DOCS] Remove legacy rollup glossary items

jrodewig:
Per elastic/elasticsearch#42720, we're refactoring rollups to phase out rollup jobs and special rollup indices.

The functionality may be around for a while, but I'd like to remove these items from the glossary. They're pulled from ES `master` where I'm removing the source defs. Once the refactor is finished, rollup jobs/indices are not something new users are likely to run into or something we want to promote.

If wanted, I can hold off on merging until closer to launch.

Relates to elastic/elasticsearch#65519


lcawl (approved):
LGTM, thanks!

PR #1528, opened 2021-01-19
Add known issue for zero memory on Debian 8

droberts195:
Relates to elastic/elasticsearch#66885

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
@@ -195,14 +195,14 @@ DELETE temp_ml_config
 
 Where possible, {ml} jobs are assigned to nodes based on the memory requirement
 of the job and the memory available on the node. However, in certain cases, the
-amount of memory on a node cannot be accurately determined, and then jobs are
-assigned by balancing numbers of jobs per {ml} node. It may lead to a situation
-where all the jobs with high memory requirements are on one node and the less
+amount of memory on a node cannot be accurately determined and jobs are assigned
+by balancing the number of jobs per {ml} node. It may lead to a situation where
+all the jobs with high memory requirements are on one node and the less
 memory-intensive jobs on another.
 
 One particular case of this problem is that {es} fails to determine the amount
-of memory on a machine that is running Debian 8 with the default Cgroups setup,
-and certain updates of Java versions earlier than Java 15.  For example, Java
+of memory on a machine that is running Debian 8 with the default Cgroups setup
+and certain updates of Java versions earlier than Java 15. For example, Java
 8u271 is known to be affected while Java 8u272 is not. Java 15 was fixed from
 its initial release.
 
@@ -210,22 +210,17 @@ If you are running {es} on Debian 8 with an old version of Java and have not
 already modified the Cgroups setup then it is recommended to do one of the
 following:
 
-1. Upgrade Java to version 15.
-2. Upgrade to the latest Java update for the version of Java you are running.
-3. Enable the "memory" Cgroup by editing `/etc/default/grub` and adding:
+* Upgrade Java to version 15.
+* Upgrade to the latest Java update for the version of Java you are running.
+* Enable the "memory" Cgroup by editing `/etc/default/grub` and adding:
 +
 --
-```
+[source,sh]
+--------------------------------------------------
 GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1"
-```
+--------------------------------------------------
+
+Update your GRUB configuration by running `sudo update-grub`, then reboot the
+machine.
+
 --
-+
-Then update your GRUB configuration by running:
-+
---
-```
-sudo update-grub
-```
---
-+
-Reboot the machine.
--- a/docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
@@ -194,23 +194,23 @@ DELETE temp_ml_config
 == Suboptimal job assignment on Debian 8
 
 Where possible {ml} jobs are assigned to nodes based on the memory requirement
-of the job and the memory available on the node. This breaks down when the
+of the job and the memory available on the node. However, in certain cases, the
 amount of memory on a node cannot be accurately determined, and then jobs are
-assigned simply by balancing numbers of jobs per {ml} node. This could lead to
-a situation where all the big jobs are on one node and all the small jobs on
+assigned by balancing numbers of jobs per {ml} node. It may lead to
+a situation where all the jobs with high memory requirements are on one node and the less memory-intensive jobs on
 another.
 
-Elasticsearch will fail to determine the amount of memory on a machine that is
+{es} fails to determine the amount of memory on a machine that is
 running Debian 8 with the default Cgroups setup, and certain updates of Java
 versions earlier than Java 15.  For example, Java 8u271 is known to be affected
-while Java 8u272 is not.  Java 15 was fixed from its initial release.
+while Java 8u272 is not. Java 15 was fixed from its initial release.
 
-If you are running Elasticsearch on Debian 8 with an old version of Java and
-have not already modified the Cgroups setup then you should do one of the
+If you are running {es} on Debian 8 with an old version of Java and
+have not already modified the Cgroups setup then it is recommended to do one of the
 following:
 
-1. Upgrade Java to version 15
-2. Upgrade to the latest Java update for the version of Java you are running
+1. Upgrade Java to version 15.
+2. Upgrade to the latest Java update for the version of Java you are running.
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,40 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+== Suboptimal job assignment on Debian 8
+
+Where possible {ml} jobs are assigned to nodes based on the memory requirement
+of the job and the memory available on the node. This breaks down when the
+amount of memory on a node cannot be accurately determined, and then jobs are
+assigned simply by balancing numbers of jobs per {ml} node. This could lead to
+a situation where all the big jobs are on one node and all the small jobs on
+another.
+
+Elasticsearch will fail to determine the amount of memory on a machine that is
```
szabosteve, line 203:
```suggestion
{es} fails to determine the amount of memory on a machine that is
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,40 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+running Debian 8 with the default Cgroups setup, and certain updates of Java
+versions earlier than Java 15.  For example, Java 8u271 is known to be affected
+while Java 8u272 is not.  Java 15 was fixed from its initial release.
```
szabosteve, line 206:
```suggestion
while Java 8u272 is not. Java 15 was fixed from its initial release.
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,40 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+
+If you are running Elasticsearch on Debian 8 with an old version of Java and
```
szabosteve, line 208:
```suggestion
If you are running {es} on Debian 8 with an old version of Java and
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,40 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+have not already modified the Cgroups setup then you should do one of the
```
szabosteve, line 209:
```suggestion
have not already modified the Cgroups setup then it is recommended to do one of the
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,40 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+following:
+
+1. Upgrade Java to version 15
```
szabosteve, line 212:
```suggestion
1. Upgrade Java to version 15.
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,40 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+2. Upgrade to the latest Java update for the version of Java you are running
```
szabosteve, line 213:
```suggestion
2. Upgrade to the latest Java update for the version of Java you are running.
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,40 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+[discrete]
+[[ml-debian8-memory]]
+== Suboptimal job assignment on Debian 8
+
+Where possible {ml} jobs are assigned to nodes based on the memory requirement
+of the job and the memory available on the node. This breaks down when the
+amount of memory on a node cannot be accurately determined, and then jobs are
+assigned simply by balancing numbers of jobs per {ml} node. This could lead to
+a situation where all the big jobs are on one node and all the small jobs on
+another.
```
szabosteve, line 201: Nit. Please feel free to ignore my suggestion if you think it does not fit. I wanted to specify a bit more accurately what "big" and "small" mean. 

```suggestion
of the job and the memory available on the node. However, in certain cases, the
amount of memory on a node cannot be accurately determined, and then jobs are
assigned by balancing numbers of jobs per {ml} node. It may lead to
a situation where all the jobs with high memory requirements are on one node and the less memory-intensive jobs on
another.
```

szabosteve (approved):
LGTM! I left a couple of suggestions, only smaller changes. Please take or leave them.

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,39 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
 DELETE temp_ml_config
 --------------------------------------------------
 // TEST[skip:TBD]
+
+[discrete]
+[[ml-debian8-memory]]
+== Suboptimal job assignment on Debian 8
+
+Where possible {ml} jobs are assigned to nodes based on the memory requirement
```
lcawl, line 196: I read this as "...possible ML jobs...", which didn't make sense, so I suggest adding a comma here:
```suggestion
Where possible, {ml} jobs are assigned to nodes based on the memory requirement
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,39 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+of the job and the memory available on the node. However, in certain cases, the
+amount of memory on a node cannot be accurately determined, and then jobs are
+assigned by balancing numbers of jobs per {ml} node. It may lead to
+a situation where all the jobs with high memory requirements are on one node and the less memory-intensive jobs on
+another.
+
+{es} fails to determine the amount of memory on a machine that is
```
lcawl, line 203: I found the connection between the first and second paragraph to be unclear. Are these Debian 8 scenarios the only "certain cases" referred to in the first paragraph, or just one example of that? I think it would help to start the second paragraph with "For example,..." or "In particular,..." or something like that to bridge the gap.

> {es} fails

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,39 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+while Java 8u272 is not. Java 15 was fixed from its initial release.
+
+If you are running {es} on Debian 8 with an old version of Java and
+have not already modified the Cgroups setup then it is recommended to do one of the
+following:
+
+1. Upgrade Java to version 15.
+2. Upgrade to the latest Java update for the version of Java you are running.
+3. Enable the "memory" Cgroup by editing `/etc/default/grub` and adding:
+
```
lcawl, line 215: This command is not being indented properly, so I think you need to add a continuation section like this:
```suggestion
+
--
```

docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc
```diff
@@ -188,3 +188,39 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+```
+GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1"
+```
+
+Then update your GRUB configuration by running:
+
+```
+sudo update-grub
+```
```
lcawl, line 224: If you add the continuation section, you need to close it:
```suggestion
```
--
```

File: docs/en/stack/ml/anomaly-detection/ml-troubleshooting.asciidoc [hunk 12]
```diff
@@ -188,3 +188,39 @@ POST _ml/set_upgrade_mode?enabled=false&timeout=10m
...
+
+And finally reboot the machine.
```
lcawl, line 226: Not a big deal, but I'd put this in the ordered list too:
```suggestion
4.  Reboot the machine.
```

droberts195: Although I originally committed this, on re-reading more carefully I remembered this is part of step 3.  You only need to choose any 1 from steps 1, 2 and 3.  I have tried to add `+` and `--` markings to bind the whole of step 3 together better.

But is it against the Elastic style to put "pick any one" options in numbered lists?  Should these three options be in a bulleted list instead?

lcawl: Oh wow, I totally skimmed over the "do one of the following...".  I agree they should be a bulleted list instead of a numbered list to make that more obvious.

PR #1571, opened 2021-02-10
Remove stray character from release highlights text

karenzone:


lcawl (approved):
Thanks for the cleanup!

PR #1605, opened 2021-03-09
[DOCS] Add lists for overloaded terms

lcawl:
For cases where a glossary term has multiple definitions, it can be unclear where one definition begins and another ends. This PR therefore adds ordered lists, which are implemented using the method described here: https://docs.asciidoctor.org/asciidoc/latest/lists/continuation/#dropping-the-principal-text

### Preview

https://stack-docs_1605.docs-preview.app.elstc.co/guide/en/elastic-stack-glossary/master/terms.html#c-glos

### DIFF
--- a/docs/en/glossary/glossary.asciidoc
+++ b/docs/en/glossary/glossary.asciidoc
@@ -117,11 +117,12 @@ Used to send {filebeat} and {metricbeat} information to the logging cluster.
 //Source: Cloud
 
 [[glossary-ml-bucket]] bucket::
+. {blank}
 +
 --
 include::{kib-repo-dir}/glossary.asciidoc[tag=bucket-def]
-
-The {ml-features} also use the concept of a bucket to divide the time series
+--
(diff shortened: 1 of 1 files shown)


gchaps (approved):
Overall, LGTM. The term "bucket" has two different meanings, I think, and should also use this format.

The description for "document" is weird.  It looks like there are two paras that say the same thing, but in slightly different words.

Not for this PR, but I think overall the glossary needs a clean up.  Some of the descriptions include details beyond defining the term, which imo, are not needed.

lcawl:
> Overall, LGTM. The term "bucket" has two different meanings, I think, and should also use this format.

Good catch!  I've updated that term's list.

> The description for "document" is weird. It looks like there are two paras that say the same thing, but in slightly different words.

I thought it was multiple definitions too, but it's all coming from the same place, so I'll follow up in the source within the elasticsearch repo.

> Not for this PR, but I think overall the glossary needs a clean up. Some of the descriptions include details beyond defining the term, which imo, are not needed.

Agreed

PR #1606, opened 2021-03-10
[DOCS] Update get-started-docker.asciidoc

flo076:
Use the correct compose file for shutdown the stack.


flo076:
"Author of some commits did not sign a contributor agreement" , it's ok now.

PR #1471, opened 2020-11-23
[DOCS] Reviews anomaly detection limitations

szabosteve:
## Overview

This PR:
* arranges the anomaly detection related limitations into four groups (and adds a short explanation about them to the intro):
    * Platform
    * Configuration
    * Operation
    * UI
* comments out `Jobs close on the datafeed end date` as the described behavior is not a limitation. (This information will be published in the AD feature documentation in a separate PR, then removed from this piece.)
* removes the following limitations:
    * Jobs must be stopped before updates (changed in 6.3)
    * Date nanoseconds data types are not supported (changed in 7.7)
* amends the following limitations:
    * Security integration
    * Anomaly Explorer and Single Metric Viewer omissions and limitations
    * Jobs created in Kibana use model plot config and pre-aggregated data.

(This PR **does not** remove `Machine learning objects do not belong to Kibana spaces` due to backporting considerations. That limitation will be modified in another PR.)

### Preview

[AD limitations](https://stack-docs_1471.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-limitations.html)

PR #1607, opened 2021-03-10
[DOCS] Reorganizes DFA limitations

szabosteve:
## Overview

This PR:
* arranges the DFA related limitations into four groups (and adds a short explanation about them to the intro):
    * Platform
    * Configuration
    * Operation
    * UI
* changes inference trained model terminology to trained model terminology where applies.

This PR is part of an effort that aims to make the AD, DFA, and Transforms limitations sections similar to each other in structure. Related PR: https://github.com/elastic/stack-docs/pull/1471

### Preview

[DFA limitations](https://stack-docs_1607.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-dfa-limitations.html)

### DIFF
--- a/docs/en/stack/ml/df-analytics/ml-dfa-limitations.asciidoc
+++ b/docs/en/stack/ml/df-analytics/ml-dfa-limitations.asciidoc
@@ -139,13 +139,12 @@ this field is skipped during the analysis.
 === {oldetection-cap} field types
 
 {oldetection-cap} requires numeric or boolean data to analyze. The algorithms 
-don't support missing values (see also <<dfa-missing-fields-limitations>>), 
-therefore fields that have data types other than numeric or boolean are ignored. 
-Documents where included fields contain missing values, null values, or an array 
-are also ignored. Therefore a destination index may contain documents that don't 
-have an {olscore}. These documents are still reindexed from the source index to the 
-destination index, but they are not included in the {oldetection} analysis and 
-therefore no {olscore} is computed.
+don't support missing values, therefore fields that have data types other than 
+numeric or boolean are ignored. Documents where included fields contain missing 
+values, null values, or an array are also ignored. Therefore a destination 
+index may contain documents that don't have an {olscore}. These documents are 
+still reindexed from the source index to the destination index, but they are not 
+included in the {oldetection} analysis and therefore no {olscore} is computed.
 
 [discrete]
 [[dfa-regression-field-type-docs-limitations]]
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/df-analytics/ml-dfa-limitations.asciidoc
```diff
@@ -147,46 +178,40 @@ majority to minority class labels in the training data. If your training data
...
+[discrete]
 [[dfa-feature-importance-limitation]]
-== Analytics runtime performance may significantly slow down with feature importance computation
+=== Analytics runtime performance may significantly slow down with {feat-imp} computation
 
 For complex models (such as those with many deep trees), the calculation of 
-feature importance takes significantly more time. Feature importance is 
-calculated at the end of the analysis and therefore the job may appear to be 
-stuck at 99% for several hours.
+{feat-imp} takes significantly more time. {feat-imp-cap} is calculated at the
```
dimitris-athanasiou, line 196: Calculation of feature importance is now done as results are written out and thus the computation progress is captured in the `writing_results` phase progress. We should no longer see jobs stuck at 99%.

docs/en/stack/ml/df-analytics/ml-dfa-limitations.asciidoc
```diff
@@ -83,28 +125,17 @@ If any of these conditions applies, the destination index of the
...
-`training_percent` to a value that represents less than 2^32^ documents.
-
-
-[float]
+[discrete]
 [[dfa-missing-fields-limitations]]
-== Documents with missing values in analyzed fields are skipped
+=== Documents with missing values in analyzed fields are skipped
 
 If there are missing values in feature fields (fields that are subjects of the
```
dimitris-athanasiou, line 132: I think we need to rephrase this a bit. Since it was added, the change is that we now support missing values depending on the analysis. So, `classification` and `regression` support missing values while `outlier_detection` does not.

We mention this just after, where we talk about limitations per analysis.

It might make more sense to remove this point and replace it with one that explains we skip rows where analyzed fields contain a multi-element array as a value. This limitation holds across analysis types.

Regarding missing values, it is probably best to only document it in the relevant section that is in the context of a specific analysis type (like we already do).

dimitris-athanasiou (commented):
Looks good. Just a couple of comments.

dimitris-athanasiou (approved):
LGTM

PR #1619, opened 2021-03-17
Update stack-docs readme

lcawl:
This PR augments the stack-docs readme (based on content in https://github.com/elastic/kibana/blob/master/README.md and https://github.com/elastic/docs/blob/master/README.asciidoc). It also adds the missing license file.

Similar PRs will be opened for the https://github.com/elastic/security-docs, https://github.com/elastic/observability-docs, and other solution-level doc-only repos.


tomcallahan (changes requested):
Elastic License v2 makes sense to me, but we shouldn't put it in a folder in this repo. We should just have it as a root-level LICENSE.txt.

lcawl:
> We should just have it as a root-level LICENSE.txt.

Thanks, Tom!  I've removed the licenses folder in https://github.com/elastic/stack-docs/pull/1619/commits/30c4967f66b83a5bbf46e250e7c030bbb9bbbea0

legalastic:
@pchang7092 adding this to get your thoughts on license to apply to a docs-only repo (e.g., ELv2 versus CC-BY versus simply holding the copyright)

@lcawl @tomcallahan @skearns64 -- to what extent are we willing to allow recipients to redistribute the docs and/or to modify and distribute modified versions?

pchang7092:
Am I understanding correctly that, while this repo contains only docs, that it contains enough of the attributes of "software" that it would be considered software within the plain meaning of the word? That for example it contains instructions that would be loaded and executed by the computer?

lcawl:
> ... That for example it contains instructions that would be loaded and executed by the computer?

All of the solution-level docs-only repos contain examples, if that's what you mean. For example:

- Stack: https://www.elastic.co/guide/en/elastic-stack-get-started/current/get-started-elastic-stack.html#install-elasticsearch
- ML: https://www.elastic.co/guide/en/machine-learning/current/ml-configuring-transform.html
- Security: https://www.elastic.co/guide/en/security/current/signals-index-template.html
- Observability: https://www.elastic.co/guide/en/observability/current/ingest-logs.html

However, these examples must be copy and pasted by users. They are not loaded and executed automatically.

pchang7092:
Thanks @lcawl . I should have clarified, I meant in terms of the way people would use this repository to access its content. I ask because the Elastic License 2.0 only covers "software". This categorical restriction is most evident in the limitations section, where EL 2.0 prevents a licensee from offering the software as a hosted service with a substantial set of the features of the software, or from circumventing the license key functionality. My question is whether this docs-only repo has enough of the traditional attributes of "software" for it to make sense to be covered by the EL 2.0 license in the same way that normal source code or object code would.

pchang7092:
I've added this item to the open source sync tomorrow and invited Lisa to the call.

pchang7092:
@lcawl , @legalastic and I met last week to discuss. Two issues surfaced in that discussion.

(1) We do not presently know the business's preferences with respect to how documentation should be licensed. In particular we were unsure whether we wanted a license that is less permissive and does not allow for re-publication - the approach taken in our main /docs [repo](https://github.com/elastic/docs) today ("The documents produced by this build process may be published only on https://www.elastic.co. They may not be published in any other form or on any other website without explicit permission.") -  or a free and open model of documentation, in which the licensee may freely use, modify, and distribute the documentation, including publish it. Disallowing publication might reduce the number of inaccurate or outdated versions of our documentation in the wild, but we also see some merit to allowing documentation to be made as permissive as the underlying software. We also recognize it may be possible to recreate some of our documentation by lifting comments from the software source code itself, which might be permitted under the applicable license terms, e.g. EL 2.0.

(2) We also felt that it would be good to be a bit more precise with respect to our license choices for documentation. Some licenses expressly contemplate documentation - e.g., MIT and Apache 2.0 address a license to documentation in one form or another - but others speak only of software concepts (e.g., Elastic License 2.0). If we're asking generally about what license is best, we can make sure the license we choose actually cleanly covers documentation so there's no interpretive gap.

@skearns64 we'd value your input on this. Would you be willing to join the open-source sync (Thursdays at 4 pm Eastern/1 pm Pacific) at your next available opportunity to discuss with the group?

PR #1621, opened 2021-03-18
Activate include for beats breaking changes

dedemorton:
Activates the include for Beats BCs.  Doc build will fail until  https://github.com/elastic/beats/pull/24646 is merged and backported.

@lcawl In situations where there are no breaking changes, I don't think we should send users down the rabbit hole, so I've removed the para that begins "This list summarizes the most important breaking changes..." WDYT? Good idea? Bad idea?

This approach only works if there really are *no* changes (even in the changelog). I think that's the case for Beats in 7.12 (I'll know for sure when 24646 is reviewed.)


lcawl:
That sounds reasonable to me!  I think we've removed that sentence in the past for the same reasons.

dedemorton:

PR #1634, opened 2021-04-14
[DOCS] Clarifies that custom rules are called job rules in Kibana

szabosteve:
## Overview

This PR adds a mention of custom rules are called job rules in Kibana to the `Custom rules` section.

### Preview

[Custom rules](https://stack-docs_1634.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-rules.html)

### NOTE

Do not merge this PR before https://github.com/elastic/kibana/pull/96999


lcawl (approved):
LGTM. Should we add that qualification in the glossary entry too?

szabosteve:
> LGTM. Should we add that qualification in the glossary entry too?

@lcawl I think we should. However, I'd like to backport this PR to 7.13. As the glossary is not version controlled, it is easier to open a separate PR than editing the backport PR.

PR #1645, opened 2021-04-22
Update create-certs.yml with "--keep-ca-key"

brucekk248:
"--keep-ca-key" helps the newbies to tackle the stack related problems like adding new host to the stack/scaling the stack/expanding the stack. I wasted weeks in finding this solution.


brucekk248:
added --keep-ca-key to the create-cert.yml. This argument helps in the long run if the person following the document want to scale his infrastructure.

PR #1650, opened 2021-04-26
Update example output to show recent versions

dedemorton:
The example shows very old lucene versions.

This should be backported to 7.13 and later.


lcawl (approved):
LGTM, thanks!

PR #1651, opened 2021-04-28
[DOCS] Adds item about fields containing arrays to anomaly detection limitations

szabosteve:
## Overview

This PR adds an item to the anomaly detection limitation section that explains what happens if the analyzed field contains an array.

### Preview

[Arrays in analyzed fields are turned into comma-separated strings](https://stack-docs_1651.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-limitations.html#ml-arrays-limitations)

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/ml-limitations.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/ml-limitations.asciidoc
@@ -75,13 +75,20 @@ also applies to those properties when you create advanced jobs in {kib}.
 [[ml-arrays-limitations]]
 === Arrays in analyzed fields are turned into comma-separated strings
 
-If an {anomaly-job} is configured to analyze a field that contains an array, 
-then the array is turned into a comma-separated concatenated string. The items 
-in the array are sorted alphabetically and the duplicated items are removed. For 
+If an {anomaly-job} is configured to analyze an aggregatable field (a field that 
+is part of the index mapping definition), and this field contains an array, then 
+the array is turned into a comma-separated concatenated string. The items in the 
+array are sorted alphabetically and the duplicated items are removed. For 
 example, the array `["zebra", "dog", "cat", "alligator", "cat"]` becomes 
-`alligator,cat,dog,zebra`. The Anomaly Explorer charts does not display results 
-as the string does not exist in the source data. The Single Metric Viewer 
-display results if the model plot is enabled.
+`alligator,cat,dog,zebra`. The Anomaly Explorer charts don't display results as 
+the string does not exist in the source data. The Single Metric Viewer displays 
+results if the model plot is enabled.
+
+If an array field is not aggregatable and is retrieved from `_source`, the array 
+is also turned into a comma-separated, concatenated list. However, the list 
+items are not sorted alphabetically, nor are they deduplicated. Taken the 
+example above, the comma-separated list, in this case, would be
+`zebra,dog,cat,alligator,cat`.
 
 Analyzing large arrays results in long strings which may require more system 
 resources. Consider using a query in the {dfeed} that filters on the relevant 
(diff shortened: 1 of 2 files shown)


docs/en/stack/ml/anomaly-detection/ml-limitations.asciidoc
```diff
@@ -70,6 +70,24 @@ You cannot use the following field names in the `by_field_name` or
...
+
+[discrete]
+[[ml-arrays-limitations]]
+=== Arrays in analyzed fields are turned into comma-separated strings
+
+If an {anomaly-job} is configured to analyze a field that contains an array, 
+then the array is turned into a comma-separated concatenated string. The items 
+in the array are sorted alphabetically and the duplicated items are removed. For 
+example, the array `["zebra", "dog", "cat", "alligator", "cat"]` becomes 
+`alligator,cat,dog,zebra`. The Anomaly Explorer charts does not display results
```
droberts195, line 82: This is actually a bit of a nightmare.

From my reading of the code in `SourceField` and `DocValueField`, my understanding is that _if_ an array field is aggregatable, i.e. has mappings that imply doc values will be stored, then what is described will happen: the distinct values will be put into the comma separated list in alphabetical order.  However, if an array field is not aggregatable and is retrieved from `_source` then they won't.  So in the example given the comma separated string would be `zebra,dog,cat,alligator,cat`.

Do you agree @dimitris-athanasiou?

dimitris-athanasiou: I have just tested this and it is exactly as you describe it.

droberts195: 😬 

I guess we have to document what happens today in the short term.  But maybe in the longer term we should make both cases work the same.

szabosteve: Thanks for the clarification and the test! @droberts195 I modified the documentation accordingly via 
c354f86.

droberts195 (approved):
LGTM

docs/en/stack/ml/anomaly-detection/ml-limitations.asciidoc
```diff
@@ -70,6 +70,31 @@ You cannot use the following field names in the `by_field_name` or
...
+the array is turned into a comma-separated concatenated string. The items in the 
+array are sorted alphabetically and the duplicated items are removed. For 
+example, the array `["zebra", "dog", "cat", "alligator", "cat"]` becomes 
+`alligator,cat,dog,zebra`. The Anomaly Explorer charts don't display results as 
+the string does not exist in the source data. The Single Metric Viewer displays 
+results if the model plot is enabled.
+
+If an array field is not aggregatable and is retrieved from `_source`, the array 
+is also turned into a comma-separated, concatenated list. However, the list 
+items are not sorted alphabetically, nor are they deduplicated. Taken the
```
lcawl, line 89:
```suggestion
items are not sorted alphabetically, nor are they deduplicated. Taking the 
```

docs/en/stack/ml/anomaly-detection/ml-limitations.asciidoc
```diff
@@ -70,6 +70,31 @@ You cannot use the following field names in the `by_field_name` or
...
+[discrete]
+[[ml-arrays-limitations]]
+=== Arrays in analyzed fields are turned into comma-separated strings
+
+If an {anomaly-job} is configured to analyze an aggregatable field (a field that 
+is part of the index mapping definition), and this field contains an array, then 
+the array is turned into a comma-separated concatenated string. The items in the 
+array are sorted alphabetically and the duplicated items are removed. For 
+example, the array `["zebra", "dog", "cat", "alligator", "cat"]` becomes 
+`alligator,cat,dog,zebra`. The Anomaly Explorer charts don't display results as
```
lcawl, line 83: > The Anomaly Explorer charts don't display results as 

Does it only omit results related to this field (e.g. in the Anomalies table) or does its presence affect the ability to render *all* results?

szabosteve: @lcawl Addressed via 5b7954c
I tested it, the charts are blank.

lcawl (approved):
LGTM, thanks!

PR #1654, opened 2021-05-03
Update glossary.asciidoc

dedemorton:
A different syntax is required to use attributes in IDs. I think it's OK to hard code.


bmorelli25 (approved):
Thanks!

PR #1659, opened 2021-05-12
[DOCS] Adds autoscaling info to Anomaly detection at scale

szabosteve:
## Overview

This PR adds some details about how to avoid jobs being tucked in "opening" state forever by using autoscaling (ESS) or setting up ML settings properly (on-prem).

### Preview

[Anomaly detection at scale – Node sizing](https://stack-docs_1659.docs-preview.app.elstc.co/guide/en/machine-learning/master/anomaly-detection-scale.html#node-sizing)

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
@@ -24,7 +24,7 @@ require you to clone an existing job or create a new one.
 
 [discrete]
 [[node-sizing]]
-== 1. Consider node sizing and configuration
+== 1. Consider node sizing/autoscaling and configuration
 
 An {anomaly-job} runs on a single node and requires sufficient resources to hold 
 its model in memory. When a job is opened, it will be placed on the node with 
@@ -54,6 +54,16 @@ Increasing the number of nodes will allow distribution of job processing as well
 as fault tolerance. If running many jobs, even small memory ones, then consider 
 increasing the number of nodes in your environment.
 
+In {ecloud}, you can define policies that scales up or down your {ml} nodes 
+based on node requirements by using {ref}/xpack-autoscaling.html[autoscaling]. 
+Set `xpack.ml.max_ml_node_size` to define the maximum possible size of a {ml} 
+node. When you try to create a {anomaly-job} that is larger than the maximum 
+size a node can support, it instantly fails as autoscaling cannot add a node big 
+enough to run the job. On a self-managed deployment, make sure that 
+`xpack.ml.max_model_memory_limit` is set according to the available resources of 
+the {ml} node. In this case, if you have a job with a larger model, it instantly 
+fails otherwise it cannot be opened due to lack of resources and it remains in 
+the `opening` state.
 
 [discrete]
 [[dedicated-results-index]]
--- a/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
@@ -54,16 +54,19 @@ Increasing the number of nodes will allow distribution of job processing as well
 as fault tolerance. If running many jobs, even small memory ones, then consider 
 increasing the number of nodes in your environment.
 
-In {ecloud}, you can define policies that scale your {ml} nodes up or down
-based on node requirements by using {ref}/xpack-autoscaling.html[autoscaling]. 
-Set `xpack.ml.max_ml_node_size` to define the maximum possible size of a {ml} 
-node. If you try to create an {anomaly-job} that is larger than the maximum 
-size a node can support, it instantly fails as autoscaling cannot add a node big 
-enough to run the job. On a self-managed deployment, make sure that 
-`xpack.ml.max_model_memory_limit` is set according to the available resources of 
-the {ml} node. In this case, if you have a job with a larger model, it instantly 
-fails. Otherwise, it cannot be opened due to lack of resources and it remains in 
-the `opening` state.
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -24,7 +24,7 @@ require you to clone an existing job or create a new one.
 
 [discrete]
 [[node-sizing]]
-== 1. Consider node sizing and configuration
+== 1. Consider node sizing/autoscaling and configuration
```
lcawl, line 27: IMO these terms should all be "ands" instead of an "and" and an "or":
```suggestion
== 1. Consider autoscaling, node sizing, and configuration
```

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -54,6 +54,16 @@ Increasing the number of nodes will allow distribution of job processing as well
 as fault tolerance. If running many jobs, even small memory ones, then consider 
 increasing the number of nodes in your environment.
 
+In {ecloud}, you can define policies that scales up or down your {ml} nodes
```
lcawl, line 57:
```suggestion
In {ecloud}, you can define policies that scale your {ml} nodes up or down
```

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -54,6 +54,16 @@ Increasing the number of nodes will allow distribution of job processing as well
 as fault tolerance. If running many jobs, even small memory ones, then consider 
 increasing the number of nodes in your environment.
 
+In {ecloud}, you can define policies that scales up or down your {ml} nodes 
+based on node requirements by using {ref}/xpack-autoscaling.html[autoscaling]. 
+Set `xpack.ml.max_ml_node_size` to define the maximum possible size of a {ml} 
+node. When you try to create a {anomaly-job} that is larger than the maximum
```
lcawl, line 60:
```suggestion
node. When you try to create an {anomaly-job} that is larger than the maximum 
```

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -54,6 +54,16 @@ Increasing the number of nodes will allow distribution of job processing as well
...
 
+In {ecloud}, you can define policies that scales up or down your {ml} nodes 
+based on node requirements by using {ref}/xpack-autoscaling.html[autoscaling]. 
+Set `xpack.ml.max_ml_node_size` to define the maximum possible size of a {ml} 
+node. When you try to create a {anomaly-job} that is larger than the maximum 
+size a node can support, it instantly fails as autoscaling cannot add a node big 
+enough to run the job. On a self-managed deployment, make sure that 
+`xpack.ml.max_model_memory_limit` is set according to the available resources of 
+the {ml} node. In this case, if you have a job with a larger model, it instantly 
+fails otherwise it cannot be opened due to lack of resources and it remains in
```
lcawl, line 65: IMO we need a semi-colon or period here:
```suggestion
fails. Otherwise, it cannot be opened due to lack of resources and it remains in 
```

lcawl (approved):
Minor suggestions, otherwise LGTM

docs/en/stack/ml/anomaly-detection/anomaly-detection-scale.asciidoc
```diff
@@ -54,6 +54,16 @@ Increasing the number of nodes will allow distribution of job processing as well
...
+In {ecloud}, you can define policies that scale your {ml} nodes up or down
+based on node requirements by using {ref}/xpack-autoscaling.html[autoscaling]. 
+Set `xpack.ml.max_ml_node_size` to define the maximum possible size of a {ml} 
+node. If you try to create an {anomaly-job} that is larger than the maximum 
+size a node can support, it instantly fails as autoscaling cannot add a node big 
+enough to run the job. On a self-managed deployment, make sure that 
+`xpack.ml.max_model_memory_limit` is set according to the available resources of 
+the {ml} node. In this case, if you have a job with a larger model, it instantly 
+fails. Otherwise, it cannot be opened due to lack of resources and it remains in 
+the `opening` state.
```
droberts195, line 66: We shouldn't recommend setting `xpack.ml.max_ml_node_size` in ESS, as the ESS infrastructure sets it to the correct value. Also, it's not compulsory to set `xpack.ml.max_model_memory_limit` in a self-managed cluster. The UI will still never suggest a model memory limit that's too high to be opened in the cluster even if this setting is not set. That setting was primarily added as a way to protect Cloud nodes against people creating jobs that were too big, but Cloud already sets this setting itself. Saying "make sure" makes it sound like this is something that's essential for people to do when it's really not in most cases.

```suggestion
In {ecloud}, you can enable {ref}/xpack-autoscaling.html[autoscaling] so that the
{ml} nodes in your cluster will scale up or down based on current {ml} memory
requirements. The {ecloud} infrastructure will allow you to create {ml-jobs} up
to the size that would fit on the maximum node size that the cluster can scale to
(usually somewhere between 58GB and 64GB) rather than what would fit in the
current cluster. If you are attempting to use autoscaling outside of {ecloud} then
set `xpack.ml.max_ml_node_size` to define the maximum possible size of a {ml} 
node. Creating {ml-jobs} with model memory limits larger than the maximum the
maximum node size can support will not be allowed, as autoscaling cannot add
a node big enough to run the job. On a self-managed deployment, you can set 
`xpack.ml.max_model_memory_limit` according to the available resources of 
the {ml} node. This will stop you creating jobs with model memory limits too high
to open in your cluster.
```

PR #1664, opened 2021-05-13
[DOCS] Add `alias` glossary term

jrodewig:
Adds the `alias` glossary term. Removes the `index alias` term.

Depends on https://github.com/elastic/elasticsearch/pull/73065


jrodewig:

PR #1682, opened 2021-06-02
[DOCS] Expands anomaly explorer and single metric viewer limitation

szabosteve:
## Overview

This PR expands the anomaly detection limitation about Anomaly Explorer and Single Metrics Viewer with the behavior in case of differing values of datafeed aggregation interval and bucket span.

Related to https://github.com/elastic/ml-team/issues/549

### Preview

[Anomaly Explorer and Single Metric Viewer limitations](https://stack-docs_1682.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-limitations.html#_anomaly_explorer_and_single_metric_viewer_omissions_and_limitations)


docs/en/stack/ml/anomaly-detection/ml-limitations.asciidoc
```diff
@@ -234,6 +234,13 @@ Explorer** nor the **Single Metric Viewer** can plot and display an anomaly
 chart for the job. In these cases, the charts are not visible and an explanatory 
 message is shown.
 
+When the aggregation interval of the {dfeed} and the bucket span of the job
+don't match, the values of the chart plotted in both the **Single Metric
+Viewer** and the **Anomaly Explorer** differ from the actual values of the job.
+To avoid this behavior, make sure that the aggregation interval in the {dfeed}
+configuration and the bucket span in the {anomaly-job} configuration have the
+same values.
```
lcawl, line 242: > same values.

I think we need to include a "However, ..." and mention why that isn't the recommendation from the back-end point of view.

szabosteve: @lcawl I added the other side of the story via 30d6c79.

lcawl (approved):
LGTM

sophiec20:
The way the limitation is worded does not make sense in a visualisation limitations section. Please alter this para to something along the lines of the following. This is a charting limitation and mentioning the datafeed aggregation interval  is likely to add to confusion:

`When viewing long time ranges in the Single Metric Viewer, the charting aggregation interval may adjust to be larger than the bucket span. This is an in-built adjustment due to limits on the total number of data points supported for each chart. When viewing long time ranges, the anomaly hover info will continue to be reported according to the job bucket span, whereas the chart will plot more aggregated values. For this reason, the Single Metric Viewer chart is best viewed with a chart aggregation interval equal to the bucket span of the job. Click on "Auto" to set the zoom width automatically to the bucket span.`

Note: The impact differs depending on sum/metric/count and if model plot is enabled or not. Going into this level of detail does not add much value to the explanation imho, as the advice is always to view the chart at the correct zoom level when you are investigating the detail. We did not limit to UI to _only_ be able to view at bucket_span granularity because navigation to the interesting point in time is made easier by being able to see the big picture and it often helps to view both zoomed in and zoomed out.

sophiec20 (changes requested):
pls see prev comment

lcawl:
> The way the limitation is worded does not make sense in a visualisation limitations section...

I had a quick chat with @sophiec20 and I think the text she provided addresses the other type of problem scenario (i.e. we'll need to describe both scenarios). She suggested that this content belongs in a "The charts in AE and SMV don't quite look like I expect" type of section and have several reason why this might occur.  If that doesn't fit in with the layout of the limitations page, perhaps we could add it to the "troubleshooting" page instead.

She also had another recommendation: `If your datafeed is an aggregation, then the charting interval you choose must also be a multiple of the datafeed aggregation interval`... though we didn't go into what would happen if that's not the case.

szabosteve:
@lcawl @sophiec20 
Thanks for the input regarding the different intervals.
If I understand correctly, we have two scenarios:
* datafeed aggregation interval and bucket span need match in order to have correct values in the charts (my PR addresses this scenario)
* charting aggregation interval and bucket span need match in order to have correct values in the charts (Sophie's comment describes this scenario)

We have to cover both scenarios. 
This PR adds its content to a section called "Anomaly Explorer and Single Metric Viewer omissions and limitations" which describes similar cases (i.e. when charts seem odd or are blank). I think we could add the second scenario (Sophie's suggestion) to this PR (and under the same limitation), so we have these cases in one place at least. Once we have it documented, we can turn it into troubleshooting content with troubleshooting steps and screenshots. WDYT?

szabosteve:
I close this PR as we shouldn't advise people to change fundamental parts of their job config in order to view charts properly.

Issue #1698, opened 2021-06-13
[404 not found]  Well, that about wraps it up for this page


jihyunjeongme:
Hello elastic ML team,

I've found that the link, `Getting started -> To get started with Machine Learning please have a look at` is 404 not found.
- https://www.elastic.co/guide/en/elastic-stack-overview/current/ml-getting-started.htm

![image](https://user-images.githubusercontent.com/43984584/121805127-bdfa9f00-cc84-11eb-802f-14d3fd28c3ad.png)

Could you check the link?

Best,
Jihyun Jeong

sophiec20:
Transferring to docs repo.

Latest Getting Started pages can be found here https://www.elastic.co/guide/en/machine-learning/current/ml-getting-started.html

The stack overview pages are listed as legacy documentation and we stopped contributing to these in 7.4; however does this page require redirects? @lcawl

lcawl:
@jihyunjeongme Thanks for contacting us! If possible, can you tell me how you reached that URL? That content doesn't exist in the "Stack Overview" even going back to its beginning (version 6.3), so if there are out-dated links in the documentation, I haven't succeeded in finding them.

jihyunjeongme:
@lcawl Thank you for the quick response!

I found that the link, here, `https://github.com/elastic/ml-cpp` > `Getting started -> To get started with Machine Learning please have a look at`
![image](https://user-images.githubusercontent.com/43984584/122202850-65234480-ced8-11eb-81f6-c8021b5ae15d.png)

lcawl:
Thank you very much!  I've created https://github.com/elastic/ml-cpp/pull/1915 to fix all those musty links

PR #1701, opened 2021-06-15
[DOCS] Add legacy header for Stack Overview

lcawl:
Related to https://github.com/elastic/stack-docs/issues/1698

This PR adds a warning at the top of pages in the Stack Overview, which indicates that this is legacy documentation.


docs/en/stack/page_header.html
```diff
@@ -0,0 +1,3 @@
+<strong>IMPORTANT</strong>This documentation is no longer maintained and might be removed.
```
lcawl, line 1:
```suggestion
<strong>IMPORTANT</strong> This documentation is no longer maintained and might be removed.
```

PR #1754, opened 2021-07-12
[DOCS] Clarifies behavior of datafeeds with or without an end time

szabosteve:
## Overview

Explains the effects of stopping a datafeed with and without an end date on the corresponding anomaly detection job.

### Preview

[Datafeeds]() --available soon


lcawl (approved):
oof, good catch! LGTM

PR #1759, opened 2021-07-21
[DOCS] Make example consistent by using .env file

wyarde:
Updated the first example to make use of the .env file to paramterize the version. This I think makes it easier to move on to the second example where this principle is applied.


szabosteve:
Closing this due to unresolved conflicts.

PR #1761, opened 2021-07-21
Specify env vars consistently via yaml dict syntax

wyarde:
Making environment variable specification consistent between the kibana service and elastic search services by using yaml dict everywhere


szabosteve:
Closing now due to unresolved conflicts before the docs freeze starts.

PR #1781, opened 2021-08-10
[7.x][DOCS] Adds a note about a workaround if DFA model is too large to fit into JVM

szabosteve:
This PR backports the changes of the following commit to the 7.x branch:
[DOCS] Adds a note about a workaround if DFA model is too large to fit into JVM #1778


docs/en/stack/ml/df-analytics/dfa-regression.asciidoc
```diff
@@ -71,6 +71,10 @@ predictions are combined.
 {regression-cap} works as a batch analysis. If new data comes into your index, 
 you must restart the {dfanalytics-job}.
```
szabosteve, line 72:
```suggestion
```

PR #1783, opened 2021-08-12
[DOCS] Reviews regression docs

szabosteve:
## Overview

This PR:
* reviews and shortens regression docs text
* cut out a section about loss functions from the regression conceptual docs and add it to the advanced concepts section
* changes internal links
* amends index file and main section file


szabosteve:
@elasticmachine test this please

szabosteve:
@elasticmachine test this please

docs/en/stack/ml/df-analytics/ml-dfa-regression.asciidoc
```diff
@@ -7,264 +7,167 @@
...
-[discrete]
-[[ml-dfanalytics-regression-evaluation]]
-=== {regression-cap} evaluation
-
-include::ml-dfa-shared.asciidoc[tag=dfa-evaluation-intro]
-
-This evaluation type is suitable for evaluating {regression} models. The 
-{regression} evaluation type offers the following metrics to evaluate the model 
-performance:
+The {regression} evaluation type offers the following metrics to evaluate the
```
lcawl, line 141: > The {regression} evaluation type offers the following metrics to evaluate the ...

I'm not sure there's enough detail in the separate "Loss functions for regression..." page to justify a separate page.  I suggest either (A) merging that content into this page, or (B) augmenting the content in that separate page (and covering all the evaluation types not just the loss ones). It seems to me that you need to have advanced/specific data science skills to know which evaluation types to use.

szabosteve: @lcawl I think the issue here is that the loss functions in the separate page are not just for evaluating model performance but they have the power to influence model training by setting their parameters (for example delta in the case of Huber). This is why I thought it would be nice to mention them here in the model performance section and then describe them on a separate page in more detail where we explain when to use them and for what. The content on this page just explains their roles in the evaluation process, the description on the separate page goes beyond it and describes them as functions, not just passive values. For this reason, I still think there is value in keeping the regression loss functions in a separate page under Advanced concepts.

szabosteve (commented):
Thank you for your review @lcawl! I addressed your feedback via e590f14.
There is only one open question about loss functions, please see my comment about it [here](https://github.com/elastic/stack-docs/pull/1783/files#r690225999) and let me know what you think.

docs/en/stack/ml/df-analytics/ml-dfa-regression.asciidoc
lcawl: So from re-reading that page, it seems like loss functions are used during model-building by all types of analytics and this can't be changed? Therefore likely not too necessary for anyone except the most curious advanced users to know about (if at all).
For regression analysis in particular, you can set it when you create the job, but it's recommended to use default values unless you're an advanced user.  
I can see why you're wanting to put this in a separate page, I'm just not sure there's enough info to justify it at this point (especially since it only really seems to be of interest to the regression use case if I'm reading that right). We'll definitely want to get a feel for when something gets split off into a separate page in this new model, but we can keep tweaking this as we figure that out with more content.

szabosteve: @lcawl Thanks for your comment!

> I can see why you're wanting to put this in a separate page, I'm just not sure there's enough info to justify it at this point

I thought loss functions could be a separate page as we already have a stand-alone page under Advanced concepts with similarly short content `Hyperparameter optimization`. I used it as precedence. I don't think the concept of loss functions is easier to grasp than hyperparameter optimization. And the tuning of these functions is similarly only for users who are deeply familiar with them. This is why I think it is better to have them separately on another page, so they won't confuse the rest of the users. However, I see the point that you've made above. Do you think it would be better to add the function descriptions back?

lcawl: I think just go with what you've got for now and we can keep pondering and revising as we get more comfortable with our divisions of content!

PR #1784, opened 2021-08-13
Provide more context for breaking changes links

karenzone:
Our breaking changes content in the [Installation and Upgrade Guide](https://www.elastic.co/guide/en/elastic-stack/master/index.html) could use a bit more context around the links.  

Screen Shot 2021-08-13 at 3 25 35 PM

At first glance, this link appears to be self-referential, when in fact, it's sending the user to the Logstash Reference. I made a proposal for Logstash (and for other books as well) to add a clue that the link will send you to another book.  

#### You get to decide for your deliverables
Like your links they way they are now? Prefer a different approach? Let me know, and we can make the change in this PR, or reverse the change for your docs. 

#### Another thought 
The breaking changes for Logstash are identical between the Installation and Upgrade Guide and the Logstash Reference. Is this statement true for other products?  

I'm considering changing the text to something like "The list of breaking changes is also available in the [Logstash Reference](https://www.elastic.co/guide/en/logstash/current/index.html)" to avoid sending users away for more info when there isn't any additional info.

### DIFF
--- a/docs/en/install-upgrade/breaking.asciidoc
+++ b/docs/en/install-upgrade/breaking.asciidoc
@@ -14,9 +14,7 @@ use and make the necessary changes so your code is compatible with {version}.
 [IMPORTANT]
 ===============================
 * Make sure you check the breaking changes for each point release up to the desired new version.
-* If you are using {ml} {dfeeds} that contain discontinued search or query
-domain specific language (DSL), the upgrade will fail. In 5.6.5 and later, the
-Upgrade Assistant provides information about which {dfeeds} need to be updated.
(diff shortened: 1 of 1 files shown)


karenzone:
@lcawl I've tagged you for initial discussion. 
`@everyone` I welcome your input as well -- now or later.

lcawl:
Theoretically the addition of book names is applicable only until we migrate (and no longer have books, hooray). But then again, I'm also hoping the elastic.co/new project will render these repeated highlights and breaking changes unnecessary by then, so let's do whatever helps here now. LGTM

lcawl (approved):

karenzone:
I went down the same thought path. We new docs system and new release notes initiative should solve these problems, but that's no reason not to mitigate them for the present. Thanks!

szabosteve:
@karenzone Is it okay if I close this PR for now?

PR #1813, opened 2021-09-10
[DOCS] Add security:authentication jobs

lcawl:
Relates to https://github.com/elastic/kibana/pull/101840 and https://github.com/elastic/stack-docs/issues/1794

### Preview

https://stack-docs_1813.docs-preview.app.elstc.co/guide/en/security/master/prebuilt-ml-jobs.html#security-authentication

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
@@ -538,7 +538,7 @@ These configurations are only available if data exists that matches the
 recognizer query specified in the
 https://github.com/elastic/kibana/blob/{branch}/x-pack/plugins/ml/server/models/data_recognizer/modules/security_auth/manifest.json#L8[manifest file].
 
-If you create these jobs in the *:ml-app:* app, you must first create the
+If you create these jobs in the *{ml-app}* app, you must first create the
 recommended
 https://github.com/elastic/kibana/blob/{branch}/x-pack/plugins/ml/server/models/data_recognizer/modules/security_auth/manifest.json#L7[{data-source}].
 
@@ -548,11 +548,11 @@ Looks for an unusually large spike in successful authentication events. This can
 be due to password spraying, user enumeration or brute force activity.
 
 Job details:::
-////
+
 * Detects anomalies where the number of events is unusually high and ignores
 cases where the count is zero (using the
 {ml-docs}/ml-count-functions.html#ml-nonzero-count[`high_non_zero_count` function]).
-////
+
 Required {beats} or {agent} integrations:::
 
 * {elastic-endpoint} integration
@@ -566,13 +566,13 @@ auth_high_count_logon_events_for_a_source_ip::
 Looks for an unusually large spike in successful authentication events events
 from a particular source IP address. This can be due to password spraying, user
 enumeration or brute force activity.
-////
+
 Job details:::
  
 * Detects anomalies where the number of events by source IP is unusually high
 and ignores cases where the count is zero (using the
 {ml-docs}/ml-count-functions.html#ml-nonzero-count[`high_non_zero_count` function]).
-////
+
 Required {beats} or {agent} integrations:::
 
 * {elastic-endpoint} integration
@@ -589,13 +589,13 @@ auth_high_count_logon_fails::
 Looks for an unusually large spike in authentication failure events. This can be
 due to password spraying, user enumeration or brute force activity and may be a
 precursor to account takeover or credentialed access.
-////
+
 Job details:::
  
 * Detects anomalies where the number of events is unusually high and ignores
 cases where the count is zero (using the
 {ml-docs}/ml-count-functions.html#ml-nonzero-count[`high_non_zero_count` function]).
-////
+
 Required {beats} or {agent} integrations:::
 
 * {elastic-endpoint} integration
@@ -610,12 +610,12 @@ Looks for a user logging in at a time of day that is unusual for the user. This
 can be due to credentialed access via a compromised account when the user and
 the threat actor are in different time zones. In addition, unauthorized user
 activity often takes place during non-business hours.
-////
+
 Job details:::
  
-* Detects anomalies where events by user happen at unusual times (using the
+* Detects anomalies where events happen at unusual times for a user (using the
 {ml-docs}/ml-time-functions.html#ml-time-of-day[`time_of_day` function]).
-////
+
 Required {beats} or {agent} integrations:::
 
 * {elastic-endpoint} integration
@@ -634,13 +634,12 @@ This can be due to credentialed access via a compromised account when the user
 and the threat actor are in different locations. An unusual source IP address
 for a username could also be due to lateral movement when a compromised account
 is used to pivot between hosts.
-////
+
 Job details:::
  
-//* Detects anomalies where the number of events is unusually high and ignores
-//cases where the count is zero
-//(using the {ml-docs}/ml-count-functions.html#ml-nonzero-count[`high_non_zero_count` function]).
-////
+* Detects unusually rare `source.ip` values compared to other users (using the
+{ml-docs}/ml-rare-functions.html#ml-rare[`rare` function]).
+
 Required {beats} or {agent} integrations:::
 
 * {elastic-endpoint} integration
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
```diff
@@ -528,6 +528,153 @@ Required ECS fields when not using {beats}:::
...
+
+* {elastic-endpoint} integration
+
+Required ECS fields:::
+
+* `event.category`
+* `event.outcome`
+
+auth_high_count_logon_events_for_a_source_ip::
+Looks for an unusually large spike in successful authentication events events
```
peteharverson, line 566: Duplicate `events` here

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
```diff
@@ -528,6 +528,153 @@ Required ECS fields when not using {beats}:::
...
+auth_rare_source_ip_for_a_user::
+Looks for a user logging in from an IP address that is unusual for the user.
+This can be due to credentialed access via a compromised account when the user
+and the threat actor are in different locations. An unusual source IP address
+for a username could also be due to lateral movement when a compromised account
+is used to pivot between hosts.
+
+Job details:::
+ 
+* Detects unusually rare `source.ip` values compared to other users (using the
```
peteharverson, line 640: This doesn't sound right to me. Isn't it detecting rare `source.ip` values that are rare for that user (rather than comparing it to other users)? The description above looks like a better description `Looks for a user logging in from an IP address that is unusual for the user.`

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
```diff
@@ -528,6 +528,153 @@ Required ECS fields when not using {beats}:::
 
 // end::siem-auditbeat-auth-jobs[]
 
+[discrete]
+[[security-authentication]]
+== Security: Authentication
+
+Detect anomalous activity in your ECS-compatible authentication logs.
+
+These configurations are only available if data exists that matches the
```
peteharverson, line 537: @spong - is this correct? Does the security app run the query in the `manifest.json` file to decide whether the displays the job configurations inside the security app?

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
```diff
@@ -528,6 +528,153 @@ Required ECS fields when not using {beats}:::
...
+[[security-authentication]]
+== Security: Authentication
+
+Detect anomalous activity in your ECS-compatible authentication logs.
+
+These configurations are only available if data exists that matches the 
+recognizer query specified in the
+https://github.com/elastic/kibana/blob/{branch}/x-pack/plugins/ml/server/models/data_recognizer/modules/security_auth/manifest.json#L8[manifest file].
+
+If you create these jobs in the *{ml-app}* app, you must first create the
```
peteharverson, line 541: This statement (and it is repeated below) is not correct. The ML UI runs the query in the `manifest.json` against the index pattern (or 'data view' from 8.0) to decide whether to display the card in the job wizard (so actually the statement you have for the security app is the one to use when discussing whether the jobs appear in the ML app).

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
```diff
@@ -854,11 +1001,15 @@ Required ECS fields:::
...
 == Security: Network
 
+Detect anomalous network activity in your ECS-compatible network logs.
+
 These configurations are only available if data exists that matches the 
 recognizer query specified in the
 https://github.com/elastic/kibana/blob/{branch}/x-pack/plugins/ml/server/models/data_recognizer/modules/security_network/manifest.json[manifest file].
 
-Detect anomalous network activity in your ECS-compatible network logs.
+If you create these jobs in the *{ml-app}* app, you must first create the
```
peteharverson, line 1010: As above, this is not correct for the ML app - we run the query in the `manifest.json` against the select index pattern / data view.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
lcawl: Thanks!  I've created https://github.com/elastic/kibana/pull/112135 to fix that in the module too.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
lcawl: I've taken another run at this statement in 33d2596b3801340540626e42ecab5a4bd5748e7c

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
lcawl: Ditto re changes in 33d2596b3801340540626e42ecab5a4bd5748e7c

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
lcawl: Since this content is used in both the ML docs and the Security docs, if this statement only applies to the ML app we can qualify it as such.

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
lcawl: Thanks!  I've updated this bullet to match what we use in the job wizard (2a5182f0b6da14999925859f524576dbce610095)

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
peteharverson: This change stating how you create the jobs in the ML app LGTM

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
peteharverson: This change stating how you create the jobs in the ML app LGTM

peteharverson (approved):
Latest edits LGTM

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
spong: @peteharverson -- we don't run the query from the manifest, but we do hit the `recognize` API providing the configured `securitySolution:defaultIndex` index pattern from Kibana Advanced Settings. Does this result in the same behavior as running the query specified in the `manifest.json`?

https://github.com/elastic/kibana/blob/4584a8b570402aa07832cf3e5b520e5d2cfa7166/x-pack/plugins/security_solution/public/common/components/ml_popover/api.ts#L23-L42

spong (approved):
Updates LGTM! 

Added a comment around how security uses the `recognize` API to determine supported modules -- if @peteharverson verifies this is the same behavior as running the query in the module's `manifest.json` then I think the existing language is good. 👍

docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
peteharverson: Yes, hitting the `recognize` API with the configured index pattern results in the same behavior as running the query specified in the `manifest.json` file. But @lcawl do you think it's worth clarifying how the check in the security app uses the index pattern configured under `securitySolution:defaultIndex` from Kibana Advance Settings?

Something like

```
These configurations are only available if data in the securitySolution:defaultIndex index pattern configured under Kibana Advanced Settings exists that matches the query specified in the manifest file.
```

lcawl: To cover both circumstances, how about something like this:

> In the Machine Learning app, these configurations are available only when data exists
that matches the query specified in the [link to manifest file]. In the Elastic Security app, it looks
in the data view specified in the [securitySolution:defaultIndex advanced setting] for data that matches the query.

peteharverson: Suggested text LGTM

lcawl: Thanks, I've added that info in e4cade7d5a93b5251aed9ff44dc91691ff5f3c3b

PR #1839, opened 2021-09-23
[DOCS] Add intro; move tutorial

lcawl:
Moves the "getting started" tutorial under "Anomaly Detection", since it does not cover data frame analytics.  
Moves the introduction from the first page of the tutorial into a new page (aligned with similar "What is...?" pages for [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/elasticsearch-intro.html), [Kibana](https://www.elastic.co/guide/en/kibana/current/introduction.html), [Observability](https://www.elastic.co/guide/en/observability/current/observability-introduction.html), etc).

### Preview

* https://stack-docs_1839.docs-preview.app.elstc.co/guide/en/machine-learning/master/machine-learning-intro.html
* https://stack-docs_1839.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-getting-started.html

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/index.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/index.asciidoc
@@ -10,6 +10,8 @@ include::ml-api-quickref.asciidoc[leveloffset=+1]
 
 include::anomaly-examples.asciidoc[leveloffset=+1]
 
+include::../get-started/index.asciidoc[leveloffset=+2]
+
 include::{es-repo-dir}/ml/anomaly-detection/ml-configuring-alerts.asciidoc[leveloffset=+2]
 
 include::{es-repo-dir}/ml/anomaly-detection/ml-configuring-aggregations.asciidoc[leveloffset=+2]
--- a/docs/en/stack/ml/machine-learning-intro.asciidoc
+++ b/docs/en/stack/ml/machine-learning-intro.asciidoc
@@ -21,7 +21,7 @@ probability model and can run continuously to identify unusual events as they
 occur. The model evolves over time; you can use its insights to forecast future
 behavior.
 
-<<ml-dfa-finding-outliers,{oldetection-cap}>> does not require time series data; 
+<<ml-dfa-finding-outliers,{oldetection-cap}>> is a type of <<ml-dfanalytics,{dfanalytics}>> that does not require time series data; 
 it identifies unusual points in a data set by analyzing how close each data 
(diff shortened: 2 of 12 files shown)


docs/en/stack/ml/anomaly-detection/index.asciidoc
```diff
@@ -10,6 +10,8 @@ include::ml-api-quickref.asciidoc[leveloffset=+1]
 
 include::anomaly-examples.asciidoc[leveloffset=+1]
 
+include::../get-started/index.asciidoc[leveloffset=+2]
```
szabosteve, line 13: I'm not sure that the Tutorial should be a subsection of Examples. What do you think about publishing it in the first level of the Anomaly detection book?

docs/en/stack/ml/machine-learning-intro.asciidoc
```diff
@@ -0,0 +1,49 @@
...
+There are two types of analysis that can deduce the patterns and relationships
+within your data without training or intervention: _{anomaly-detect}_ and
+_{oldetection}_.
+
+<<ml-ad-overview,{anomaly-detect-cap}>> requires time series data. It constructs a
+probability model and can run continuously to identify unusual events as they
+occur. The model evolves over time; you can use its insights to forecast future
+behavior.
+
+<<ml-dfa-finding-outliers,{oldetection-cap}>> does not require time series data;
```
szabosteve, line 24: I think it's a great idea to split the different types of analyses in terms of being supervised or unsupervised. However, we have another layer that we use to organize ML content and the TOC reflects that logic (anomaly detection/data frame analytics). For this reason, I think we should mention in the short description of outlier detection, classification, and regression that those are different types of data frame analytics. Something like:
```suggestion
<<ml-dfa-finding-outliers,{oldetection-cap}>> is a type of <<ml-dfanalytics,{dfanalytics}>> that does not require time series data; 
```
What do you think?

szabosteve (commented):
Thank you for these changes, I think the new structure and intro represent the ML features much better! I have two questions, otherwise LGTM!

docs/en/stack/ml/anomaly-detection/index.asciidoc
lcawl: IMO it is less important now that the "Finding anomalies" is our intended end-to-end description. So if not in the examples, I'd likely embed it under that page.  For now, I'll leave in examples, since I still need to revisit the location of all the items in that section.

PR #1851, opened 2021-10-19
Change - {version} by ${VERSION}

rauhmaru:
Adjust in config files, changing {version} by ${VERSION}


gtback:
cla/check

gtback:
Thanks @rauhmaru ! I'm not a Docker compose expert so I'll leave this to @bytebilly or @lockewritesdocs to review. It looks like in #1906 this file got changed to having CRLF line endings. I'm not sure if this was intentional, but it took a bit of work to resolve the conflict of getting the latest `main` merged into this.

bytebilly (changes requested):
Uh, this change doesn't work as it should be `${STACK_VERSION}` to correctly load the variable from the `.env` file.
This file should be exactly the same as `https://github.com/elastic/elasticsearch/tree/master/docs/reference/setup/install/docker-compose.yml`.

Is there a way to ensure that we have just one version of it? This will prevent future desync.

gtback:
I went looking into how this file was used, and it seems like it's substituting AsciiDoc variables in [here](https://github.com/elastic/stack-docs/blob/main/docs/en/getting-started/get-started-docker.asciidoc#L397) rather than variables from the `.env` file. That's why it shows "8.0.0" [here](https://www.elastic.co/guide/en/elastic-stack-get-started/current/get-started-stack-docker.html#docker-compose-file). 

In terms of single-sourcing the file, we don't currently depend on the `elasticsearch` repo in the Getting Started guide, but we could add that around [here](https://github.com/elastic/docs/blob/master/conf.yaml#L206-L216). It's probably best to let Adam take a look when he's back.

bytebilly:
> I went looking into how this file was used, and it seems like it's substituting AsciiDoc variables in [here](https://github.com/elastic/stack-docs/blob/main/docs/en/getting-started/get-started-docker.asciidoc#L397) rather than variables from the .env file. That's why it shows "8.0.0" [here](https://www.elastic.co/guide/en/elastic-stack-get-started/current/get-started-stack-docker.html#docker-compose-file).

Yes, that's the problem that I tried to fix in https://github.com/elastic/elasticsearch/pull/83823, but I haven't realized that docs use a different source file.

I'd suggest to start fixing the docs in this PR (by just copying the right file in this repo), and then we can iterate and see if we can have a ssot.

lcawl:
I have updated the Getting Started Guide to pull the docker-compose.yml .env files from the elasticsearch repo in https://github.com/elastic/stack-docs/pull/2069, so I believe this PR can be closed.  Thank you for drawing attention to this discrepancy!

gtback:
Thanks for bringing this to our attention, @rauhmaru ! Even though we ended up not merging this, it prompted us to improve this documentation, and we are very grateful for that! 🙏🏻

PR #1858, opened 2021-10-25
[DOCS] Updates ML setup page

szabosteve:
## Overview

This PR updates the `Set up machine learning features` page.

### Preview

[Set up ML](https://stack-docs_1858.docs-preview.app.elstc.co/guide/en/machine-learning/master/setup.html#requirements-overview)


sophiec20 (approved):
LGTM with the following ..

1. `{data-sources-cap}` occurs twice on page - this might just be a problem rendering in the preview.
2. Please get @pheyos to dbl check the facts
3. Could the screenshot for assigning a job to a space include the spaces icon - not sure how much it would help, but the non-intuitive bit is knowing you have to click on the icon to display the fly-out

image

Thanks for your patience in working through these changes.

docs/en/stack/ml/setup.asciidoc
```diff
@@ -5,100 +5,140 @@
...
-`xpack.ml.enabled` is `true`, however, they can service API requests. For more
-information, see {ref}/modules-node.html#ml-node[{ml-cap} nodes] and
-{ref}/ml-settings.html[{ml-cap} settings in {es}].
 
 [discrete]
 [[setup-privileges]]
 == Security privileges
 
+Security privileges can be assigned as either {es} privileges or {kib} feature 
+privileges. The user is granted the most permissive combination of both. You can
```
pheyos, line 35: Nit picking here:
`The user is granted the most permissive combination of both.` is only true when using Kibana (non dev tools). We're restricting it two sentences later: `If you use {ml-features} via {es} APIs [...]`, but I'm wondering if we want to slightly modify this initial statement to make it crystal clear.

docs/en/stack/ml/setup.asciidoc
```diff
@@ -5,100 +5,140 @@
...
-* [ ] `all` {kib} privileges for the {ml-app} or Discover features in
-the appropriate spaces. Alternatively, `read` {kib} privileges for the {ml-app}
-feature and `all` {kib} privileges for the {ipm-app} feature
-* [ ] `manage_pipeline` or `manage_ingest_pipelines` cluster privileges
-* [ ] `create`, `create_index`, `manage` and `read` index privileges for
-destination indices
+- [ ] Kibana feature privileges that grant use of File Upload, either:
+  * `All` Machine Learning or
+  * `All` Discover or
+  * `Read` Machine Learning and `All` Data View Management
```
pheyos, line 137: I've tested all of the combinations and couldn't get it working with just `All` for Discover (and ofc all required index privileges). Not sure where this requirement came from, but we should double check what to expect - maybe we have a bug here.

pheyos (commented):
Overall LGTM, just two comments.

docs/en/stack/ml/setup.asciidoc
szabosteve: Thank you for your help @pheyos! I'd like to backport these changes to 7.16. Just to double-check: does the security information apply to 7.16 as well?

pheyos: @szabosteve I think we have some small differences in 7.16, but let me double check.

docs/en/stack/ml/setup.asciidoc
```diff
@@ -5,100 +5,145 @@
...
-* [ ] `all` {kib} privileges for the {ml-app} or Discover features in
-the appropriate spaces. Alternatively, `read` {kib} privileges for the {ml-app}
-feature and `all` {kib} privileges for the {ipm-app} feature
-* [ ] `manage_pipeline` or `manage_ingest_pipelines` cluster privileges
-* [ ] `create`, `create_index`, `manage` and `read` index privileges for
-destination indices
+- [ ] `All` or `Read` Machine Learning or `All` Discover {kib} feature 
+  privileges
+- [ ] `All` {data-sources-caps} Management {kib} feature privileges (required 
+  for automatic {data-source} creation)
```
pheyos, line 142: This is actually also required to create data views via the data frame analytics wizard. In the section above we have the note `{data-sources-cap} can be automatically created when creating a 
{dfanalytics-job}.`. I'm wondering if we'd like to repeat this requirement in the note above or if we even want to extract this bit such that it work for both sections.
Any thoughts on this @szabosteve @sophiec20 ?

docs/en/stack/ml/setup.asciidoc
pheyos: The changes we had from 7.16 to 8.0 are only around ML now being part of Kibana's `base` privileges. We tell the reader that `All Kibana privileges for the machine learning features` or `Read Kibana privileges for the machine learning features` are required, but we don't mention HOW to set these, so our documentation is valid for both versions and I can't spot anything that would need a 7.16 specific change.

docs/en/stack/ml/setup.asciidoc
```diff
@@ -5,100 +5,145 @@
...
-enabled, there are some {ml} settings that can be updated only by operator 
-users.
+[%interactive]
+- [ ] the {subscriptions}[appropriate subscription] level or the free trial 
+  period activated
+- [ ] `xpack.ml.enabled` set to its default value of `true` on every node in the 
+  cluster (refer to {ref}/ml-settings.html[{ml-cap} settings in {es}])
+- [ ] `ml` value defined in the list of `node.roles` on the 
+  {ref}/modules-node.html#ml-node[{ml} nodes]
+- [ ] {ml} features visible in the {kib} space
```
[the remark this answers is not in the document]
sophiec20: This top section is meant to give summary info for all aspects; license, node settings, spaces and privileges. For now, I think it is worth mentioning all at a high level.

docs/en/stack/ml/setup.asciidoc
```diff
@@ -5,100 +5,145 @@
...
+=== {kib} user
+
+Within a {kib} space, for full access to the {ml-features}, you must have:
 
 [%interactive]
-* [ ] `all` {kib} privileges for the {ml-features} in the appropriate spaces
-* [ ] `read` and `view_index_metadata` index privileges on source indices
-* [ ] `read`, `manage`, and `index` index privileges on destination indices (for
-  {dfanalytics-jobs} only)
+* [ ] `All` {kib} privileges for the {ml-features}
```
sophiec20, line 107: It has just struck me that we can be clearer here..

`All Kibana privileges for the machine learning features` -- the words "machine learning" are a long way from "all".

Should this be:
`Machine Learning: All` Kibana feature privileges.

This might be a nice way to differentiate Kibana and es privileges.

pheyos: I like this change - makes it easier to read and also easier to match the docs text with the options in the roles UI. We should apply this style to all Kibana feature privileges that we mention here (Machine Learning, Discover, Data View Management).

szabosteve: Changed in 6796c74.

docs/en/stack/ml/setup.asciidoc
szabosteve: Changed in 6796c74.

docs/en/stack/ml/setup.asciidoc
```diff
@@ -5,100 +5,145 @@
...
 
 [%interactive]
-* [ ] `read` {kib} privileges for the {ml-features} in the appropriate spaces
-* [ ] `read` index privileges on source indices
-* [ ] `read` index privileges on destination indices (for {dfanalytics-jobs}
-only)
+* [ ] Machine Learning: `Read` {kib} privileges
+* [ ] {data-sources} for your source indices
+* [ ] `read` index privilege on your source indices
+* [ ] {data-sources} and `read` index privilegeson destination indices (for
```
pheyos, line 121: Nit: 
```suggestion
* [ ] {data-sources} and `read` index privileges on destination indices (for 
```

pheyos (commented):
Changes are looking good, just found one typo.

pheyos (approved):
LGTM

PR #1883, opened 2021-11-17
[DOCS] Updating Stack getting started for security ON

lockewritesdocs:
Updates the getting started page for the Stack to reflect changes for security ON by default. Also removes old links, updates other links, and fixes some minor issues throughout the page.

Preview link: https://stack-docs_1883.docs-preview.app.elstc.co/guide/en/elastic-stack-get-started/master/get-started-elastic-stack.html

### DIFF
--- a/docs/en/getting-started/get-started-stack.asciidoc
+++ b/docs/en/getting-started/get-started-stack.asciidoc
@@ -56,7 +56,7 @@ work with your system:
 * <<linux, linux>> for Linux
 * <<win, win>> for Windows
 
-When you start {es} for the first time, the following security configuration
+When you start {es} for the first time, security features such as authentication, authorization and network encryption (TLS) for elasticsearch are enabled by default. The following security configuration
 occurs automatically:
 
 * {ref}/configuring-stack-security.html#stack-security-certificates[Certificates and keys] for TLS are
@@ -406,7 +406,7 @@ https://www.elastic.co/guide/en/kibana/current/index.html[{kib} Reference].
 ==== Access the {kib} web interface
 
 To access the {kib} web interface, point your browser to port `5601`. For example,
-https://127.0.0.1:5601[http://127.0.0.1:5601].
+http://127.0.0.1:5601[http://127.0.0.1:5601].
 
 [float]
 [[install-beats]]
--- a/docs/en/getting-started/get-started-stack.asciidoc
+++ b/docs/en/getting-started/get-started-stack.asciidoc
@@ -1,7 +1,9 @@
 [[get-started-elastic-stack]]
 == Getting started with the {stack}
 
-Looking for an {stack} ("ELK") guide that shows how to quickly install and configure the {stack}? You're in the right place! You can install the {stack} on a single VM, or even on your laptop:
+Looking for a guide that shows how to quickly install and configure the {stack}?
+You're in the right place! You can install the {stack} on a single VM, or even
+on your laptop. Install each component in the following order:
 
 * <<install-elasticsearch,{es}>>
 * <<install-kibana,{kib}>>
@@ -14,11 +16,6 @@ isn't required. To get started with {ls}, see
 After completing the installation process, learn how to implement a system
 monitoring solution that uses {metricbeat} to collect server metrics and ship
 the data to {es}. Then use {kib} to search and visualize the data.
-
-IMPORTANT: Implementing security is a critical step in configuring the {stack}.
-This guide skips security configuration to quickly install a sample installation. Before sending sensitive data across the network,
-{ref}/elasticsearch-security.html[secure the {stack}] and enable
-{ref}/encrypting-communications.html[encrypted communications].
 
 [float]
 [[install-prereqs]]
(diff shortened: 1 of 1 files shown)


docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -43,18 +40,33 @@ data, such as logs or decoded network packets.
...
+* <<win, win>> for Windows
 
+When you start {es} for the first time, the following security configuration
+occurs automatically:
+
+* {ref}/configuring-stack-security.html#stack-security-certificates[Certificates and keys] for TLS are
+generated for the transport and HTTP layers.
+* The TLS configuration settings are written to `elasticsearch.yml`.
+* A password is generated for the `elastic` user.
+* An enrollment token is generated for {kib}.
```
jkakavas, line 66: This is not true for deb and rpm, not sure how we can differentiate it here without making it too complicated though

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -43,18 +40,33 @@ data, such as logs or decoded network packets.
...
-for Linux, and <<win, win>> for Windows):
+work with your system:
+
+* <<deb, deb>> for Debian/Ubuntu
+* <<rpm, rpm>> for Redhat/Centos/Fedora
+* <<mac, mac>> or <<brew, brew>> for OS X
+* <<linux, linux>> for Linux
+* <<win, win>> for Windows
 
+When you start {es} for the first time, the following security configuration
```
jkakavas, line 59: maybe a sentence that is more user friendly, similar to
```suggestion
When you start {es} for the first time, security features such as authentication, authorization and network encryption (TLS) for elasticsearch are enabled by default. The following security configuration
```
? Think it's important we tell users _what_ we do for them as well as how

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -200,16 +212,22 @@ To learn more about installing, configuring, and running {es}, read the
...
+Open a new terminal and verify that you can connect to your {es} cluster by
+making an authenticated call. Enter the password for the `elastic` user when
+prompted:
 
 [source,shell]
-----------------------------------------------------------------------
-curl http://127.0.0.1:9200
-----------------------------------------------------------------------
+----
+curl --cacert config/tls_auto_config_<timestamp>/http_ca.crt \
```
jkakavas, line 223: we haven't told them that they need to be in the dir where they installed elasticsearch so a relative path `config/tls_auto_config_<timestamp>/http_ca.crt` is confusing. We should either tell them to navigate to that dir or (preferably IMO) qualify the path with $ES_CONF_PATH (linking to where we describe where that points to in each installation type.

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -200,16 +212,22 @@ To learn more about installing, configuring, and running {es}, read the
...
+-u elastic https://localhost:9200 <1>
+----
+// NOTCONSOLE
+<1> Ensure that you use `https` in your call, or the request will fail.
```
jkakavas, line 227: I've seen in other places over the last PRs and it troubled me more and more so now reached my "need-to-comment" threshold :) 

The <1> callout makes sense in the context of the diff of this PR but not necessarily when you read the docs as a user. The callout is true, but it's also _too(?)_ obvious. As in we could also say, `Ensure that you use `9200` as the port in your call, or the request will fail.`

WDYT ?

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -361,10 +403,10 @@ To learn more about installing, configuring, and running {kib}, read the
 https://www.elastic.co/guide/en/kibana/current/index.html[{kib} Reference].
 
 [float]
-==== Launch the {kib} web interface
+==== Access the {kib} web interface
 
-To launch the {kib} web interface, point your browser to port 5601. For example,
-http://127.0.0.1:5601[http://127.0.0.1:5601].
+To access the {kib} web interface, point your browser to port `5601`. For example,
+https://127.0.0.1:5601[http://127.0.0.1:5601].
```
jkakavas, line 409: we're not enabling TLS for kibana by default, so this is still `http`

jkakavas (changes requested):

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -1,7 +1,9 @@
 [[get-started-elastic-stack]]
 == Getting started with the {stack}
 
-Looking for an {stack} ("ELK") guide that shows how to quickly install and configure the {stack}? You're in the right place! You can install the {stack} on a single VM, or even on your laptop:
+Looking for a guide that shows how to quickly install and configure the {stack}?
```
dedemorton, line 4: Can you check with @AnneB-SEO  to see if users are still searching for ELK tutorials? (We added this awhile ago because users were finding 3rd party instructions rather than our docs.)

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -249,20 +267,44 @@ and maps.
...
+[NOTE]
+====
+If you need to reset the password for the `elastic` user or other
+built-in users, run the {ref}/reset-password.html[`elasticsearch-reset-password`] tool.
+To generate new enrollment tokens for {kib} or {es} nodes, run the
+{ref}/create-enrollment-token.html[`elasticsearch-create-enrollment-token`] tool.
+These tools are available in the {es} `bin` directory.
+====
+
+[[deb-rpm-linux]]*deb, rpm, or linux:*
```
dedemorton, line 307: Might consider using the tabbed widgets for platform-specific commands. (I can see why you wouldn't want to go to the trouble, though.)

dedemorton (commented):
Thanks for updating the docs! This is going to be a great change for users.

docs/en/getting-started/get-started-stack.asciidoc
lockewritesdocs, line 409:
```suggestion
http://127.0.0.1:5601[http://127.0.0.1:5601].
```

docs/en/getting-started/get-started-stack.asciidoc
lockewritesdocs: Talked to Anne, and ELK > Elastic Stack. Thanks for pointing this out! I'll add "ELK" back in.

docs/en/getting-started/get-started-stack.asciidoc
lockewritesdocs: ++ I think we can change this to:

>Path to the generated `http_ca.crt` certificate for the HTTP layer. This file is created in the {es} configuration directory, which is defined by the`$ES_PATH_CONF` [environment variable](https://www.elastic.co/guide/en/elasticsearch/reference/8.0/settings.html#config-files-location).

docs/en/getting-started/get-started-stack.asciidoc
lockewritesdocs: We can add a note that indicates the difference with a link to the `elasticsearch-create-enrollment-token` tool:

>**NOTE**: On `deb` and `rpm` installations, an enrollment token isn't generated for {kib}. To generate an enrollment token for {kib}, use the [`elasticsearch-create-enrollment-token`](https://www.elastic.co/guide/en/elasticsearch/reference/8.0/create-enrollment-token.html) tool.

jkakavas: "isn't generated for {kib}." -> "isn't generated for {kib} on installation." is my only suggestion. ++ otherwise

docs/en/getting-started/get-started-stack.asciidoc
jkakavas: @lockewritesdocs  TBC, I'm not suggesting we add "Ensure that you use 9200 as the port in your call, or the request will fail." here. It was just a comment about another thing that while true, is (IMO) too obvious to point out and I doubt the usefulness. My suggestion is to remove the existing callout.

lockewritesdocs: I thought about this a bit, and I agree that we can remove this callout. If we get feedback that users get stuck, we can always add something.

PR #1886, opened 2021-11-22
[DOCS] Adds data views to glossary

gchaps:
This PR defines the term data views in the glossary.

### DIFF
--- a/docs/en/glossary/glossary.asciidoc
+++ b/docs/en/glossary/glossary.asciidoc
@@ -325,7 +325,7 @@ role] that typically share the same hardware profile. Data tiers include the
 
 [[glossary-data-view]] data view::
 An object that enables you to select the data that you want to use in {kib} 
-and allows you to define properties of the fields. A data view can point to one or more 
+and define properties of the fields. A data view can point to one or more 
 <<glossary-data-stream,data streams>>, <<glossary-index,indices>>, or
 <<glossary-alias,aliases>>. For example, a data view can point to your log data from yesterday, 
 or all indices that contain your data.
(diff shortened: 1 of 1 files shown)


docs/en/glossary/glossary.asciidoc
```diff
@@ -323,6 +323,14 @@ role] that typically share the same hardware profile. Data tiers include the
 <<glossary-frozen-tier,frozen tier>>. See {ref}/data-tiers.html[Data tiers].
 //Source: Elasticsearch
 
+[[glossary-data-view]] data view::
+An object that enables you to select the data that you want to use in {kib} 
+and allows you to define properties of the fields. A data view can point to one or more
```
lcawl, line 328: Since this sentence already contains "enables you to", IMO you don't need the "allows you to":
```suggestion
and define properties of the fields. A data view can point to one or more 
```
I also wonder if it should be "define the properties of the fields" or "define properties for the fields", but am not sure which is better here.

lcawl (approved):
One suggestion, otherwise LGTM

jrodewig (approved):
LGTM with @lcawl's suggestion.

docs/en/glossary/glossary.asciidoc
```diff
@@ -323,6 +323,14 @@ role] that typically share the same hardware profile. Data tiers include the
 <<glossary-frozen-tier,frozen tier>>. See {ref}/data-tiers.html[Data tiers].
 //Source: Elasticsearch
 
+[[glossary-data-view]] data view::
+An object that enables you to select the data that you want to use in {kib} 
+and define properties of the fields. A data view can point to one or more
```
gchaps, line 328:
```suggestion
and define the properties of the fields. A data view can point to one or more 
```

PR #1888, opened 2021-11-22
[DOCS] Update Docker getting started for the Stack

lockewritesdocs:
This PR replaces the steps for [Running the Stack with Docker Compose](https://www.elastic.co/guide/en/elastic-stack-get-started/8.0/get-started-docker.html#run-stack-docker) with steps to start the Stack on Docker using auto-configured security. 

Preview link: https://stack-docs_1888.docs-preview.app.elstc.co/guide/en/elastic-stack-get-started/master/get-started-stack-docker.html

A follow-up PR will address updating the steps for _Running the Stack with Docker Compose_.


docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,177 @@
...
-NOTE: Version {version} of {es} has not been released,
-so the sample compose file is not yet available for this version.
-See the {stack-gs-current}/get-started-docker.html[current version] for the latest sample files.
+
+WARNING: Version {version} of {es} has not yet been released, so no
+Docker image is currently available for this version.
+
 endif::[]
+
+ifeval::["{release-state}"!="unreleased"]
```
bytebilly, line 83: Should we show the command anyway, like we do for Kibana in the following section?

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,177 @@
...
-curl -X GET "localhost:9200/_cat/nodes?v&pretty"
---------------------------------------------------
+----
+curl --cacert http_ca.crt -u elastic https://localhost:9200
+----
+// NOTCONSOLE
 
-. Open {kib} to load sample data and interact with the cluster:
-http://localhost:5601.
+. In a new terminal session, start {kib} and connect it to your {es} container:
```
bytebilly, line 152: Should we add the warning in case the version is not released yet?

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,177 @@
...
++
+--
 ifeval::["{release-state}"!="unreleased"]
-["source","yaml",subs="attributes"]
---------------------------------------------
-include::docker/docker-compose.yml[]
---------------------------------------------
+[source,sh,subs="attributes"]
+----
+docker run --name es-node01 -p 9200:9200 -p 9300:9300 -it docker.elastic.co/elasticsearch/elasticsearch:{version}
```
bytebilly, line 106: Do we really need to expose port 9300 to the world since we are creating a single-node cluster?

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,177 @@
...
+If you need to reset the password for the `elastic` user or other
+built-in users, run the {ref}/reset-password.html[`elasticsearch-reset-password`]
+tool. To generate new enrollment tokens for {kib} or {es} nodes, run the
+{ref}/create-enrollment-token.html[`elasticsearch-create-enrollment-token`] tool.
+These tools are available in the {es} `bin` directory of the Docker container.
+For example:
+
+[source,sh]
+----
+docker exec -it es-node01 /usr/share/elasticsearch/bin/elasticsearch-reset-password
```
bytebilly, line 129:
```suggestion
docker exec -it es-node01 /usr/share/elasticsearch/bin/elasticsearch-reset-password -u elastic
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,177 @@
...
+. Copy the `http_ca.crt` security certificate from your Docker container to
+your local machine.
 +
-["source","sh",subs="attributes"]
---------------------------------------------
-docker-compose up
---------------------------------------------
+[source,sh]
+----
+docker cp es-node01:/usr/share/elasticsearch/config/tls_auto_config_*/http_ca.crt .
```
bytebilly, line 138: This command is not working, as `*` cannot be used for path expansion from outside docker. There are a few changes that we can do to make it simpler, otherwise we should list folders first or find another way.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,177 @@
...
+----
+// NOTCONSOLE
 
-. Open {kib} to load sample data and interact with the cluster:
-http://localhost:5601.
+. In a new terminal session, start {kib} and connect it to your {es} container:
++
+[source,sh,subs="attributes"]
+----
+docker run --name kib-01 --net elastic -p 5601:5601 docker.elastic.co/kibana/kibana:{version}
```
bytebilly, line 156: This command fails because the `elastic` network doesn't exist.

bytebilly, line 156: Container name should be consistent across Elasticsearch and Kibana, e.g. `kibana01` here or `es-01` there.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,177 @@
...
-=== Run in Docker with TLS enabled
+  .. In your browser, paste the enrollment token that you copied when starting
+  {es} and click the button to connect your {kib} instance with {es}.
 
-If {security-features} are enabled, you must configure Transport Layer Security
-(TLS) encryption for the {es} transport layer. While it is possible to use a
-trial license without setting up TLS, we advise securing your stack from the
-start.
+  .. Log in to {kib} as the `elastic` user with the password that was generated
+  when you started {es}.
```
bytebilly, line 167: We could add clean-up  instuctions to stop and remove containers after use.

bytebilly (changes requested):
@lockewritesdocs I left suggestions for improvements and I marked commands that fail if run as-is. Ping me if you need extra technical details.

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: I'll add `--net elastic` to the command for starting Elasticsearch on line 106.

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: We could just list the command as `tls_auto_config_<timestamp>` and indicate that users need to modify that portion of their command to match the filename on disk. Open to alternatives if you have preferred options.

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: Thanks for pointing out this omission -- I'll update the command with the `-u` option across all of the Docker documentation.

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: This is the approach we implemented in the docs for [installing Kibana on Docker](https://www.elastic.co/guide/en/kibana/8.0/docker.html#run-kibana-on-docker-for-dev). If we don't want to expose port `9300`, then I'll update it in both places.

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: I'll change to `es-node01` throughout (and subsequently `es-node02` and `es-node01`) and `kib01`.

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: ++ I'll add a section for removing the containers.

lockewritesdocs:
@bytebilly -- I addressed your comments and made a few other improvements that I'll port to the other Docker docs. Let me know if additional changes are needed!

docs/en/getting-started/docker/docker-compose.yml
```diff
@@ -1,13 +1,13 @@
-version: '2.2'
+version: "2.2"
```
jkakavas, line 1: IIUC, this has been optional for a while and the current version is 3.8. Do we use any features below that require us to stay in 2.2 ? Can we change to latest or remove this ? @bytebilly might know

docs/en/getting-started/docker/docker-compose.yml
```diff
@@ -1,13 +1,13 @@
...
 services:
-  es01:
+  es-node01:
     image: docker.elastic.co/elasticsearch/elasticsearch:{version}
-    container_name: es01
+    container_name: es-node01
     environment:
```
jkakavas, line 6: I suggest that if we're keeping this compose file around, we should make it obvious that TLS is not enabled, and do so by setting xpack.security.http.ssl.enabled=false and xpack.security.transport.ssl.enabled=false here. Autoconfiguration won't be triggered because of the other settings, but I suggest we make this explicit too by setting xpack.security.autoconfiguration.enabled=false

jkakavas: I didn't read the PR description so this comment is most probably relevant for the follow up PR.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
-[float]
-[[run-stack-docker]]
-=== Run with Docker Compose
+You can <<run-docker-secure,start the {stack} in Docker>> with security enabled
+and configured by default. This option is great for quickly getting started with
+{es} and {kib} for testing and development.
+
+You can also <<get-started-docker-tls,start the {stack} with Docker Compose>> to
+create a secured, multi-node cluster with a connected {kib} instance. This
+option requires more manual steps, but results in a more robust cluster.
```
jkakavas, line 16: Should we qualify what we mean with "robust cluster" here ? It reads slightly generic

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
+
+[[run-docker-secure]]
+[discrete]
+=== Start {es} and {kib} in Docker
+Starting in {es} 8.0, security is enabled by default. With security enabled,
+{stack} {security-features} require TLS encryption for the transport networking
+layer, or your cluster will fail to start.
```
jkakavas, line 23: The TLS comment makes me think as a reader that I need to so something extra because security is enabled. I think this wording made sense in 7.x but maybe not so now ? It is true, but TLS will be enable by default if users follow what we write here so maybe we shouldn't point it out in that manner in the first paragraph ?

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
+for the transport and HTTP layers.
+* The Transport Layer Security (TLS) configuration settings are written to
+`elasticsearch.yml`.
+* A password is generated for the `elastic` user.
+* An enrollment token is generated for {kib}.
+
+You can then {kibana-ref}/docker.html[start {kib}] and enter the enrollment
+token, which is valid for 30 minutes. This token automatically applies the
+security settings from your {es} cluster, authenticates to {es} with the
+`kibana_system` user, and writes the security configuration to `kibana.yml`.
```
jkakavas, line 39:
```suggestion
`kibana` service account, and writes the security configuration to `kibana.yml`.
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
+
+[discrete]
+==== Prerequisites
+. Install the appropriate https://docs.docker.com/get-docker/[Docker application]
+for your operating system.
++
+[NOTE]
+====
+Make sure that Docker is allotted at least 4GiB of memory. In Docker Desktop,
```
jkakavas, line 48: Is that still valid ? I run my docker containers with less with no issue

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
+
+[[docker-start-enroll]]
+[discrete]
+==== Start a single-node {es} cluster and enroll {kib}
+
+The following command starts a single-node {es} cluster for development or
+testing.
+
+. Start {es} in Docker. A password is generated for the `elastic` user and
+output to the terminal, plus an enrollment token for enrolling {kib}.
```
jkakavas, line 101: we generate a token for es nodes too

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
+If you're starting a single-node {es} cluster in a Docker container, security
+will be automatically enabled and configured for you. When you start {es} for
+the first time, the following security configuration occurs automatically:
+
+* {ref}/configuring-stack-security.html#stack-security-certificates[Certificates and keys] are generated
+for the transport and HTTP layers.
+* The Transport Layer Security (TLS) configuration settings are written to
+`elasticsearch.yml`.
+* A password is generated for the `elastic` user.
+* An enrollment token is generated for {kib}.
```
jkakavas, line 34:
```suggestion
* An enrollment token is generated for {kib}.
* An enrollment token is generated for other {es} nodes.
```

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
+
+WARNING: Version {version} of {es} has not yet been released, so no
+Docker image is currently available for this version.
+
+endif::[]
+
+ifeval::["{release-state}"!="unreleased"]
+[source,sh,subs="attributes"]
+----
+docker run --name es-node01 --net elastic -p 9200:9200 -it docker.elastic.co/elasticsearch/elasticsearch:{version}
```
jkakavas, line 114: if we don't expose 9300, then no other node can join this cluster. I'm not sure we want to do this, but if we do we should point it out at least.

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,69 +1,269 @@
...
+your local machine. 
 +
-["source","sh",subs="attributes"]
+[source,sh]
+----
+docker cp es-node01:/usr/share/elasticsearch/config/tls_auto_config_<timestamp>/http_ca.crt .
+----
+`<timestamp>`:: The timestamp of when the auto-configuration process created
+the security files directory in your Docker container. Use the Docker CLI to
+locate the timestamp for this directory.
```
jkakavas, line 149: We will see about changing this to be a standard dir name, but for now we can add the command for the user's reference:
```
docker exec -ti es-node01 /bin/bash -c "find /usr/share/elasticsearch -name http_ca.crt"
``

**Review by jkakavas — CHANGES_REQUESTED** [review]

File: docs/en/getting-started/docker/docker-compose.yml (hunk 8 above)
Comment (bytebilly) [reply docs/en/getting-started/docker/docker-compose.yml:1]: We are rewriting the `docker-compose.yml` file entirely, so we don't know yet.
Since this is for getting started, we should not assume users have the latest `docker-compose` and it would be nice to make it compatible with the lowest version possible.

It's better to make it explicit, since some of the keywords have been removed in version 3 but they still work if you specify version 2 in the file.

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 16 above)
Comment (bytebilly) [reply docs/en/getting-started/get-started-docker.asciidoc:114]: Actually, other nodes running on the same host can join the cluster. I'm not sure if we want to support multi-node from multiple hosts running docker in the getting started experience, as I assume that if you run with docker, you use the same host (but I may be wrong).

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 3 above)
Comment (bytebilly) [reply docs/en/getting-started/get-started-docker.asciidoc]: Thanks

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 6 above)
Comment (bytebilly) [reply docs/en/getting-started/get-started-docker.asciidoc]: I'm not sure if this solves the problem, IIRC the network must be created explicitly.
Can we just use the default network instead?

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 13 above)
Comment (lockewritesdocs) [reply docs/en/getting-started/get-started-docker.asciidoc:48]: It's definitely valid for running a multi-node cluster with Docker Compose. With less memory than this, one node in my cluster continually crashed on macOS. However, this setting might not apply to the single-node cluster. If that's the case, I'll move this note to the section for using Docker Compose.

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 14 above)
Comment (lockewritesdocs) [reply docs/en/getting-started/get-started-docker.asciidoc:102]: ++ I'll change this to read:
>A password is generated for the `elastic` user and output to the terminal, plus enrollment tokens for enrolling {kib} and adding additional nodes to your cluster.

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 17 above)
Comment (lockewritesdocs) [reply docs/en/getting-started/get-started-docker.asciidoc:150]: Fancy! Thanks for this command; I'll definitely incorporate it into the workflow.

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 13 above)
Comment (lockewritesdocs) [reply docs/en/getting-started/get-started-docker.asciidoc:49]: For a single-node cluster, I can run both the Elasticsearch container and the Kibana container with only 1.5 GB of memory, provided that no other containers are running. If I have another container running (like a local doc build) and try to start both Elasticsearch and Kibana, Docker stops one of those containers to regain memory.

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 11 above)
Comment (lockewritesdocs) [reply docs/en/getting-started/get-started-docker.asciidoc:23]: Good point. I think that we can delete the second sentence and just say:
>Starting in {es} 8.0, security is enabled and configured by default.

File: docs/en/getting-started/get-started-docker.asciidoc (hunk 10 above)
Comment (lockewritesdocs) [reply docs/en/getting-started/get-started-docker.asciidoc:16]: Yeah, I think that we can expand a bit more without getting too in depth. How about:
>This option requires more manual steps, but results in a more resilient cluster with greater capacity and reliability.

File: docs/en/getting-started/get-started-docker.asciidoc [hunk 18]
```diff
@@ -1,75 +1,300 @@
...
++
+TIP: You might need to scroll back a bit in the terminal to view the password
+and enrollment token.
 
-. Run `docker-compose` to bring up the three-node {es} cluster and {kib}:
+. Copy the generated password and enrollment token and save them in a secure
+location. These values are shown only when you start {es} for the first time.
 +
-["source","sh",subs="attributes"]
+[NOTE]
```
debadair, line 128: I think I'd pull this info about resetting the elastic password out of the procedure into a separate subsection.

lockewritesdocs: I added a new section for generating passwords and enrollment tokens and moved the relevant information there. I'll probably make this change in the Elasticsearch docs as well.

Preview link: https://stack-docs_1888.docs-preview.app.elstc.co/guide/en/elastic-stack-get-started/master/get-started-stack-docker.html#docker-generate

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,75 +1,298 @@
...
-=== Run with Docker Compose
+You can <<run-docker-secure,start the {stack} in Docker>> with security enabled
+and configured by default. This option is great for quickly getting started with
+{es} and {kib} for testing and development.
 
-To get the default distributions of {es} and {kib} up and running in Docker,
-you can use Docker Compose.
+You can also <<get-started-docker-tls,start the {stack} with Docker Compose>> to
+create a secured, multi-node cluster with a connected {kib} instance. This
+option requires more manual steps, but results in a more resilient cluster with
```
jkakavas, line 16: we're trying to make it a single step too, so maybe we don't need defensive language here anymore ?

docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -1,75 +1,298 @@
...
+. Copy the `http_ca.crt` security certificate from your Docker container to
+your local machine. 
++
+[source,sh]
+----
+docker cp es-node01:/usr/share/elasticsearch/config/tls_auto_config_<timestamp>/http_ca.crt .
+----
+`<timestamp>`:: The timestamp of when the auto-configuration process created
+the security files directory in your Docker container.
+
```
jkakavas, line 145: we will update that as we remove the timestamp from the dirname, but I haven't had the time to do this change yet. I'll make sure I'll adjust docs as part of that change

jkakavas (approved):
LGTM

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: I'll update this language in the subsequent PR that introduces the simplified Docker Compose configuration.

docs/en/getting-started/get-started-docker.asciidoc
lockewritesdocs: We'll need to make that change across the docs where it's used, but will be an easy change.

PR #1889, opened 2021-11-24
Add comment for maintainers to getting started docs

dedemorton:
Added note about keeping ELK in this doc for SEO so this info doesn't get lost in the future.


AnneB-SEO (approved):
Looks good! Elk stack, elk stack tutorials, elk tutorials, get about 4x as many searches on Google than do elastic stack, and is preferred by many search users

PR #1897, opened 2021-11-29
[DOCS] Adds highlights for Kibana

gchaps:
This PR removes the comments so that the what's new content can be pulled from the Kibana repo.


lcawl (approved):
LGTM

PR #1930, opened 2021-12-21
[DOC] explicate confidence interval on single metrics viewer

stefnestor:
👋🏼 Explicates for search discoverability purposes that our "bounds" terminology is the 95% central confidence interval

### DIFF
--- a/docs/en/stack/ml/get-started/ml-gs-results.asciidoc
+++ b/docs/en/stack/ml/get-started/ml-gs-results.asciidoc
@@ -48,10 +48,9 @@ time. It is available only if the job has `model_plot_config` enabled. It can
 display only a single time series.
 
 The blue line in the chart represents the actual data values. The shaded blue
-area represents the bounds for the expected values. The area between the upper
-and lower bounds are the most likely values for the model. If a value is outside
-of this area then it can be said to be anomalous. The shaded blue, therefore, is
-the 95% central confidence interval.
+area represents the bounds for the expected values, which is also known as the _confidence interval_. The area between the upper
+and lower bounds are the most likely values for the model, using a 95% confidence level. That is to say, there is a 95% chance of the actual value falling within these bounds. If a value is outside
+of this area then it can be said to be anomalous.
 
 If you slide the time selector from the beginning to the end of the data, you
 can see how the model improves as it processes more data. At the beginning, the
--- a/docs/en/stack/ml/get-started/ml-gs-results.asciidoc
+++ b/docs/en/stack/ml/get-started/ml-gs-results.asciidoc
@@ -48,9 +48,11 @@ time. It is available only if the job has `model_plot_config` enabled. It can
 display only a single time series.
 
 The blue line in the chart represents the actual data values. The shaded blue
-area represents the bounds for the expected values, which is also known as the _confidence interval_. The area between the upper
-and lower bounds are the most likely values for the model, using a 95% confidence level. That is to say, there is a 95% chance of the actual value falling within these bounds. If a value is outside
-of this area then it can be said to be anomalous.
+area represents the bounds for the expected values. The area between the upper
+and lower bounds are the most likely values for the model, using a 95%
+confidence level. That is to say, there is a 95% chance of the actual value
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/get-started/ml-gs-results.asciidoc
```diff
@@ -50,7 +50,8 @@ display only a single time series.
 The blue line in the chart represents the actual data values. The shaded blue
 area represents the bounds for the expected values. The area between the upper
```
lcawl, line 51: Since we're talking about the shaded area in this sentence, I think it makes sense to add the new info here too:
```suggestion
area represents the bounds for the expected values, which is also known as the _confidence interval_. The area between the upper
```

docs/en/stack/ml/get-started/ml-gs-results.asciidoc
```diff
@@ -50,7 +50,8 @@ display only a single time series.
...
 and lower bounds are the most likely values for the model. If a value is outside
```
lcawl, line 52:
```suggestion
and lower bounds are the most likely values for the model, using a 95% confidence level. That is to say, there is a 95% chance of the actual value falling within these bounds. If a value is outside
```

docs/en/stack/ml/get-started/ml-gs-results.asciidoc
```diff
@@ -50,7 +50,8 @@ display only a single time series.
...
-of this area then it can be said to be anomalous.
+of this area then it can be said to be anomalous. The shaded blue, therefore, is
```
lcawl, line 53: If my two earlier suggestions are accepted, this final sentence is no longer required:
```suggestion
of this area then it can be said to be anomalous.
```

docs/en/stack/ml/get-started/ml-gs-results.asciidoc
```diff
@@ -50,7 +50,8 @@ display only a single time series.
...
+the 95% central confidence interval.
```
lcawl, line 54: Ditto re removing sentence:
```suggestion

```

docs/en/stack/ml/get-started/ml-gs-results.asciidoc
```diff
@@ -48,8 +48,8 @@ time. It is available only if the job has `model_plot_config` enabled. It can
 display only a single time series.
 
 The blue line in the chart represents the actual data values. The shaded blue
-area represents the bounds for the expected values. The area between the upper
-and lower bounds are the most likely values for the model. If a value is outside
+area represents the bounds for the expected values, which is also known as the _confidence interval_. The area between the upper
```
peteharverson, line 51: Perhaps we should lose the `which is also known as the confidence interval` part of this sentence, as we don't use that label anywhere in the UI as far as I know, and then just leave '95% confidence level' in the next sentence.

lcawl: Agreed, I've made that change in c7d81b9678c8cf771661b913d05a4a3393e74e04

peteharverson (approved):
LGTM

docs/en/stack/ml/get-started/ml-gs-results.asciidoc
```diff
@@ -49,8 +49,10 @@ display only a single time series.
 
 The blue line in the chart represents the actual data values. The shaded blue
 area represents the bounds for the expected values. The area between the upper
-and lower bounds are the most likely values for the model. If a value is outside
-of this area then it can be said to be anomalous.
+and lower bounds are the most likely values for the model, using a 95%
+confidence level. That is to say, there is a 95% chance of the actual value
+falling within these bounds. If a value is outside of this area then it can be
+said to be anomalous.
```
tveasey, line 55: We don't always raise an anomaly in this case although usually do. Perhaps the wording here can be tweaked to avoid questions like "we didn't get an anomaly but you said we should"
```suggestion
falling within these bounds. If a value is outside of this area then it will usually
be identified as anomalous.
```

tveasey (approved):
Sorry for the delay. Just suggested a minor tweak to avoid questions about missing anomalies

PR #1934, opened 2022-01-10
[ML] Adds common problems to NLP pipelines docs and add on_failure example

davidkyle:
### Preview

https://stack-docs_1934.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-nlp-inference.html


docs/en/stack/ml/nlp/ml-nlp-inference.asciidoc
```diff
@@ -197,6 +198,10 @@ POST _reindex
 However, those web log messages are unlikely to contain enough words for the
 model to accurately identify the language.
 
+TIP: When using deployed NLP models, set the reindex `size` option to a value
+smaller than the deployment's queue size. Otherwise, requests might be rejected
```
lcawl, line 202: Is this the queue_capacity mentioned in https://www.elastic.co/guide/en/elasticsearch/reference/master/start-trained-model-deployment.html? If so, it might be clearer to use that setting name explicitly. For example:

```suggestion
smaller than the `queue_capacity` for the trained model deployment. Otherwise, requests might be rejected
```

docs/en/stack/ml/nlp/ml-nlp-inference.asciidoc
```diff
@@ -197,6 +198,10 @@ POST _reindex
 However, those web log messages are unlikely to contain enough words for the
 model to accurately identify the language.
 
+TIP: When using deployed NLP models, set the reindex `size` option to a value
```
lcawl, line 201: Since this page is entirely about NLP models, I don't think this introductory qualifier adds any value:
```suggestion
TIP: Set the reindex `size` option to a value
```

docs/en/stack/ml/nlp/ml-nlp-inference.asciidoc
```diff
@@ -111,6 +122,42 @@ Hungarian.
 .. If everything looks correct, close the panel, and click **Create
 pipeline**. The pipeline is now ready for use.
 
+////
```
lcawl, line 125: FYI I've added the failure processor into the "Add an inference processor..." step, rather than at the end. I've commented out the API example for now, since we don't currently have those in this page. However, I will circle back in a later PR to add a full API example since @sophiec20 asked for this too.

lcawl (approved):
One outstanding suggestion, otherwise LGTM

PR #1936, opened 2022-01-10
Use ess-leadin-short shared attr

DanRoscigno:
Azure was missing from the list of Elastic Cloud providers, switched existing test to use the shared attributes, and removed closing line as it is covered in the first line of the section.


kilfoyle (approved):
LGTM! 🚢 
Nice to know about that `ess-leadin-short` attribute.

PR #1944, opened 2022-01-18
Update ml-nlp-lang-ident with zxx special case

benwtrent:
Adds new documentation referencing the new `zxx` special case. 

Relates to: https://github.com/elastic/elasticsearch/pull/82746

### Preview

https://stack-docs_1944.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-nlp-classify-text.html#ml-nlp-lang-ident


lcawl:
IMO this information was clear enough that we could omit the example strings and integrate the text right into the existing paragraph, rather than creating a new subsection.  I've made this change in 8f020202e52943b3f94ef9054ef1a5ac73a868df

lcawl (approved):

PR #1951, opened 2022-01-25
[DOCS] Removes APM NodeJS and RUM Javascript modules

lcawl:
Relates to https://github.com/elastic/kibana/pull/119945

Also adds a link from the anomaly detection examples page to the examples repo, related to https://github.com/elastic/examples/pull/389

### Preview

https://stack-docs_1951.docs-preview.app.elstc.co/guide/en/machine-learning/master/ootb-ml-jobs-apm.html
https://stack-docs_1951.docs-preview.app.elstc.co/guide/en/machine-learning/master/anomaly-examples.html


peteharverson (commented):
The deletions here are good. But there are a couple of issues I spotted with the remaining content, the second one of which should definitely be addressed. 

- We now only have the one configuration that works on APM data, so I wonder if we should switch the plural references in the first paragraph `These anomaly detection job wizards appear in Kibana` etc? Note the only jobs are in the `apm_transaction` folder.
- The rest of the text needs updating to reflect the changes to the APM transaction jobs made for 8.0.0 in https://github.com/elastic/kibana/pull/117836 (see https://github.com/elastic/kibana/pull/111957 for full description as that change was reverted and only re-applies for 8.0.0).

lcawl:
> The deletions here are good. But there are a couple of issues I spotted with the remaining content, the second one of which should definitely be addressed.

Thanks for the review!  I've addressed the first suggestion in this PR and will address the second in a subsequent PR.

PR #1962, opened 2022-02-01
Temporarily blocked shared region to Logstash breaking changes

karenzone:


karenzone (commented):
Reviewed preview. It checks out.

PR #1964, opened 2022-02-01
Temporarily block shared region to LS breaking changes 8.0

karenzone:


karenzone (commented):
Reviewed preview. It checks out.

PR #1973, opened 2022-02-04
fix: minor typos on get-started-stack.asciidoc

vunderkind:
Nothing dramatic - just two typos.

### DIFF
--- a/docs/en/getting-started/get-started-stack.asciidoc
+++ b/docs/en/getting-started/get-started-stack.asciidoc
@@ -2,7 +2,7 @@
 == Getting started with the {stack}
 
 Looking for an {stack} ("ELK") guide that shows how to set up the {stack} and 
-get up and running quickly? You're on the right place! First you install the 
+get up and running quickly? You're in the right place! First you install the 
 core products:
 
 * <<install-elasticsearch,{es}>>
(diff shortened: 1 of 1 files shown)


vunderkind:
To the maintainer of this project: I signed the contributor agreement after the fact - but can't seem to be able to re-trigger the `cla-checker-service` bot — would you be so kind?

docs/en/getting-started/get-started-stack.asciidoc
```diff
@@ -441,7 +441,7 @@ data visualizations, in about 5 minutes.
 In this section, you learn how to run the `system` module to collect metrics
 from the operating system and services running on your server. The system module
 collects system-level metrics, such as CPU usage, memory, file system, disk IO,
-and network IO statistics, as well as top-like statistics for every process
+and network IO statistics, as well as top-line statistics for every process
```
lcawl, line 444: My first impression was that this was referring to the "top" command, @dedemorton would it be accurate to change to something like this?:
```suggestion
and network IO statistics, as well as top-level statistics for every process
```

dedemorton: No, it is referring to statistics similar to what you get with the top command. If you think it's too confusing, I would just delete "top-like" from the sentence.

PR #2019, opened 2022-02-09
Update ootb-ml-jobs-siem.asciidoc

dharada:
`destination.as.organization.name` ECS field seems bo be also required for `rare_destination_country` prebuilt ML job becasue a user have experienced to fail to start the prebuilt ML job with the following 400 error.

```
{
"error": {
"root_cause": [
{
"type": "status_exception",
"reason": "[datafeed-rare_destination_country] cannot retrieve field [destination.as.organization.name] because it has no mappings"
}
],
"type": "status_exception",
"reason": "[datafeed-rare_destination_country] cannot retrieve field [destination.as.organization.name] because it has no mappings"
},
"status": 400
}
```

Please double check if this PR is approperiate. Thanks much.

### DIFF
--- a/docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
+++ b/docs/en/stack/ml/anomaly-detection/ootb-ml-jobs-siem.asciidoc
@@ -1154,7 +1154,9 @@ Required ECS fields:::
 
 * `destination.as.organization.name`
 * `destination.geo.country_name`
+* `destination.ip`
 * `event.category`
+* `source.ip`
 
 // end::security-network-jobs[]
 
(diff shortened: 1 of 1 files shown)


lcawl:
Thanks!  I've added two more fields based on the details in https://github.com/elastic/kibana/blob/7.17/x-pack/plugins/ml/server/models/data_recognizer/modules/security_network/ml/datafeed_rare_destination_country.json and https://github.com/elastic/kibana/blob/7.17/x-pack/plugins/ml/server/models/data_recognizer/modules/security_network/ml/rare_destination_country.json

lcawl (approved):
LGTM, thanks!

PR #2045, opened 2022-02-14
Add mergify configuration file for stack-docs

lcawl:
The security and observability repos use Mergify to automate backports. This PR copies the method used in observability-docs (https://github.com/elastic/observability-docs/blob/main/.mergify.yml), since it uses new, backport-specific labels (backport-x.y), rather than changing/overloading the behaviour of existing labels.

Before we merge this PR, we must create backport labels that meet the conditions, as seen in https://github.com/elastic/observability-docs/labels?q=backport

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,95 @@
+pull_request_rules:
+  - name: ask to resolve conflict
+    conditions:
+      - -merged
+      - -closed
+      - conflict
+    actions:
+        comment:
+          message: |
+            This pull request is now in conflict. Could you fix it @{{author}}? 🙏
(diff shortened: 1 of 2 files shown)


gtback (approved):
I don't have a lot of experience with Mergify, but this looks good to me.

gtback:
@bmorelli25 would you mind taking a look at this?

bmorelli25 (approved):
Yay! LGTM.

You should also consider adding mergify checks to the repo. Maybe that's something you already have in the pipeline, but I know they validate the yml to avoid any mistakes. https://github.com/elastic/observability-docs/runs/5042303818

PR #2069, opened 2022-02-24
[DOCS] Pull Getting Started Docker files from elasticsearch repo

lcawl:
Relates to https://github.com/elastic/stack-docs/pull/1851
Removes the Docker configuration files from stack-docs and updates the Getting Started Guide to reference the files from the elasticsearch repo instead.

This PR must be merged after https://github.com/elastic/docs/pull/2384

NOTE: These sections are omitted from pre-release branches, so the changes will only be visible in the backport PR for 8.0.


docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -373,7 +373,7 @@ referenced by the `docker-compose.yml` file.
 
 ["source","txt",subs="attributes"]
```
gtback, line 374: I wonder if we should take out the `subs="attributes"` since I think this is what prompted the change in #1851 (since the file on its own on GitHub doesn't work without the Asciidoc templating to replace variables

gtback: Nevermind. I think we can do this for the `docker-compose.yml` file, but not the `.env` file since it still includes `{version}`.

gtback (approved):
LGTM!

PR #2080, opened 2022-03-08
Beats and Logstash version compatible with es 8.0

daverick:
Based on the [matrix compatibility](https://www.elastic.co/fr/support/matrix#matrix_compatibility) only Beats and Logstash 7.17 are compatible with 8.x not any 7.n.

If this is not the case, the matrix compatibility should be updated instead.


docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -66,7 +66,7 @@ even if you opt to do a full-cluster restart of your {es} cluster.**
 Alternatively, you can create a new {version} deployment and reindex from remote.
 For more information, see <<upgrading-reindex, Reindex to upgrade>>.
 
-Beats and Logstash 7.n are compatible with {es} {version} 
+Beats and Logstash 7.17 are compatible with {es} {version}
```
karenzone, line 69:
```suggestion
Beats and Logstash {prev-major-last} are compatible with {es} {version} 
```

debadair (approved):
LGTM with Karen's suggestion

PR #2091, opened 2022-03-30
add note to use 127.0.0.1 when connecting to Kibana docker container

DanRoscigno:
The Kibana docker image provides a URL to connect to to continue the integration with Elasticsearch.  It uses `0.0.0.0` as the `kibana.yml` specifies to listen on all interfaces.  In order to connect to the container we generally will need to use `127.0.0.1`.


debadair (approved):
LGTM

PR #2096, opened 2022-04-10
[DOCS] Update certificate path in Docker installation

jkakavas:
Update the path where auto generated certificates are stored in the docker instructions

### DIFF
--- a/docs/en/getting-started/get-started-docker.asciidoc
+++ b/docs/en/getting-started/get-started-docker.asciidoc
@@ -121,6 +121,9 @@ endif::[]
 TIP: You might need to scroll back a bit in the terminal to view the password
 and enrollment token.
 
+. Copy the generated password and enrollment token and save them in a secure
+location. These values are shown only when you start {es} for the first time.
+
 . Copy the `http_ca.crt` security certificate from your Docker container to
 your local machine. 
 +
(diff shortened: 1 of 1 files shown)


docs/en/getting-started/get-started-docker.asciidoc
```diff
@@ -121,26 +121,13 @@ endif::[]
...
-. Copy the generated password and enrollment token and save them in a secure
-location. These values are shown only when you start {es} for the first time.
-
-. Locate the CA certificate in your container. This command returns the path
-to the `http_ca.crt` certificate, which you'll use in the next step.
-+
-[source,sh]
-----
-docker exec -it es01 /bin/bash -c "find /usr/share/elasticsearch -name http_ca.crt"
-----
```
lockewritesdocs, line 133: @jkakavas, did you mean to remove these steps? I believe that users should still copy the generated password and enrollment token, and the command to locate the CA cert is useful. Let me know if this wasn't your intention, or if something was altered that necessitated this change.

jkakavas: Thanks for catching this, it was simply an over-jealous selection. I only meant to remove the ". Locate the CA cert.." step. This was only useful so that we could find the path in the container and then use it below in the `docker cp` command. Now, we _know_ the path in the container so we can reference it directly in the `docker cp` below

PR #2134, opened 2022-05-10
[DOCS] Remove threading parameter names from NLP deployment task

lcawl:
Relates to https://github.com/elastic/elasticsearch/pull/86597, which renames the model_threads and inference_threads trained model options.

This PR updates https://www.elastic.co/guide/en/machine-learning/master/ml-nlp-deploy-models.html#ml-nlp-deploy-model, such that it no longer goes into detail about these parameters but rather links to the API reference for details.

If the recommendation about the number of physical CPU cores or virtualized CPUs is still required, we can add it to the API. However, it seems like that's already covered in the API parameter descriptions.


dimitris-athanasiou (approved):
LGTM This is a good change!

PR #2156, opened 2022-06-09
Adds guide about how to revert to a model snapshot

szabosteve:
## Overview

This PR adds a guide to the anomaly detection documentation about how to revert to a model snapshot.

### Preview

[Reverting to a model snapshot](https://stack-docs_2156.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-reverting-model-snapshot.html)


docs/en/stack/ml/anomaly-detection/ml-revert-model-snapshot.asciidoc
```diff
@@ -0,0 +1,43 @@
...
+
+. In {kib}, navigate to **{ml-app} > Anomaly Detection > Jobs**.
+. Locate the {anomaly-job} whose model you want to revert in the job table.
+. Open the job details and navigate to the **Model Snapshots** tab.
++
+--
+[role="screenshot"]
+image::images/anomaly-job-model-snapshots.jpg[A screenshot of a job with the Model Snapshots tab opened]
+--
+. Select the snapshot you want to revert from the list and click the **Revert**
```
lcawl, line 19: Not mandatory, but I think this is clearer since "reverting a snapshot" implies you're doing something to the snapshot itself:
```suggestion
. Select a snapshot from the list and click the **Revert** 
```

docs/en/stack/ml/anomaly-detection/ml-revert-model-snapshot.asciidoc
```diff
@@ -0,0 +1,43 @@
...
+. Locate the {anomaly-job} whose model you want to revert in the job table.
+. Open the job details and navigate to the **Model Snapshots** tab.
++
+--
+[role="screenshot"]
+image::images/anomaly-job-model-snapshots.jpg[A screenshot of a job with the Model Snapshots tab opened]
+--
+. Select the snapshot you want to revert from the list and click the **Revert** 
+  icon under **Actions**.
+. Optional: Select if you want to replay the analysis after the revert has been
```
lcawl, line 21: > . Optional: Select if you want to replay the analysis after the revert has been 

It might be nice to clarify what is meant by "replay the analysis" since I can't tell from this quick read and scanning the screenshot what that means. Does it mean that it's re-applying analysis that it already did in the past on top of the reverted job?  That seems to contradict the statement at the bottom of the page ("In case of reverting to a snapshot, all anomaly detection results are deleted for the corresponding job after the point when the snapshot was saved.")

docs/en/stack/ml/anomaly-detection/ml-revert-model-snapshot.asciidoc
```diff
@@ -0,0 +1,43 @@
...
+  applied. You can decide whether you want the job running in real time. You can 
+  also define the time range you don't want the job to analyze, this way you can 
+  skip the time of the event you don't want to impact the model behavior. 
++
+--
+[role="screenshot"]
+image::images/revert-model-snapshot.jpg[A screenshot of a revert model snapshot flyout]
+--
+. Click **Apply**.
+
```
lcawl, line 31: I think it might also be a good idea to mention/link to custom rules in this page too. IIRC they are a way to tell the model to avoid updating the model in certain circumstances (and therefore potentially avoiding future situations where you'd need to revert).  There's a blurb about them in https://stack-docs_2156.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-ad-run-jobs.html#ml-ad-rules

szabosteve: Added in [2381b34](https://github.com/elastic/stack-docs/pull/2156/commits/2381b347b6dbb6617ed11826718ad1e2296b74b9).

docs/en/stack/ml/anomaly-detection/ml-revert-model-snapshot.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+TIP: You can use <<ml-ad-rules,custom rules>> to avoid a model being updated in 
+  case of a known event you want to exclude from the analysis. Using custom 
+  rules might help you to avoid situations where you need to revert to a 
+  snapshot.
+
+Alternatively, you can use the 
+{ref}/ml-revert-snapshot.html[Revert model snapshots] API. In this case, you
```
lcawl, line 46: This doesn't require capitalization either:
```suggestion
{ref}/ml-revert-snapshot.html[revert model snapshots] API. In this case, you 
```

docs/en/stack/ml/anomaly-detection/ml-revert-model-snapshot.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+  snapshot.
+
+Alternatively, you can use the 
+{ref}/ml-revert-snapshot.html[Revert model snapshots] API. In this case, you 
+need to manually close the corresponding job before reverting to the saved 
+snapshot.
+
+[NOTE]
+===============================
+* In case of reverting to a snapshot, all {anomaly-detect} results are deleted
```
lcawl, line 52: I found "in case of reverting to a snapshot" unclear.  Which particular case is it referring to? Only when you do it via API?  Or some other circumstance that I'm missing? If it's applicable to all methods, something like this might be clearer:
```suggestion
* By default, when you revert to a snapshot, all {anomaly-detect} results are deleted 
```

lcawl (approved):
A few minor comments, otherwise LGTM and is an interesting read!

PR #2178, opened 2022-07-12
Moves out third-party NLP model list from Resources

szabosteve:
## Overview

This PR moves out the third-party NLP models page from the `Resources` section. Currently, there is no other content/page in this section, so putting the content there only means that users need to click through the Resources section to get the page. We might want to reconsider adding back `Resources` when we'll have more content that could be added to the section.


davidkyle (approved):
LGTM

PR #2209, opened 2022-08-11
[DOCS] Adds list of supported functions to jobs-from-visualizations guide

szabosteve:
## Overview

This PR adds more details to the `How to create AD jobs from Lens visualizations` guide:
* list of supported anomaly detector functions
* unsupported charts.

### Preview

[Prerequs]()


docs/en/stack/ml/anomaly-detection/ml-jobs-from-visuals.asciidoc
```diff
@@ -15,9 +15,14 @@ You can create {anomaly-jobs} from the compatible Lens charts on **Dashboard**.
 `area`, `area_percentage_stacked`, `area_stacked`, `bar`, `bar_horizontal`, 
 `bar_horizontal_stacked`, `bar_percentage_stacked`, `bar_stacked`, and `line`.
 
+* Supported {anomaly-detect} functions are `average`, `count`, `max`, `median`, 
+`min`, `sum`, `unique_count`.
+
 * The chart must contain a date field on one axis and it must be the same as the 
 default date field for the {data-source}.
 
+* Charts containing time shifts are not supported.
```
jgowdyelastic, line 24: also per field "filter by" adjustments are now allowed.
I'm not sure of the best way to describe these.

image

docs/en/stack/ml/anomaly-detection/ml-jobs-from-visuals.asciidoc
```diff
@@ -9,15 +9,24 @@ You can create {anomaly-jobs} from the compatible Lens charts on **Dashboard**.
...
+`min`, `sum`, `unique_count`.
+
 * The chart must contain a date field on one axis and it must be the same as the 
 default date field for the {data-source}.
 
+* Charts with a layer that contains a 
+{kibana-ref}/create-a-dashboard-of-panels-with-ecommerce-data.html#compare-time-ranges[time shift] 
+are not supported.
+
+* Charts containing a layer that is filtered by a field are not supported.
```
jgowdyelastic, line 28: I think this is slightly wrong.
I think something like this might be clearer:
"Chart layers which contain a field which has a 'filter by' setting applied cannot be used to create an anomaly detection job"

It's also worth pointing out that each layer can be user to create a job. so if a chart contains two layers, one which users a filter by or time shift, the other layer could still be used.

jgowdyelastic: Same with the time shift sentence, it should say
"Chart layers which contain a field which is using a time shift...."
Maybe both sentences could be merged?

szabosteve: Thanks for the clarification @jgowdyelastic! I updated the PR via [ec0fe99](https://github.com/elastic/stack-docs/pull/2209/commits/ec0fe99c089dc3e4af9462e99e92c0c91f79e537).

jgowdyelastic (approved):
LGTM

PR #2245, opened 2022-11-04
[DOCS] Add update API link to deployment NLP steps

lcawl:
Relates to https://github.com/elastic/elasticsearch/pull/90728

This PR adds a link to the new [update trained model deployment API](https://www.elastic.co/guide/en/elasticsearch/reference/master/update-trained-model-deployment.html) in the [Deploy NLP trained model steps](https://www.elastic.co/guide/en/machine-learning/master/ml-nlp-deploy-models.html#ml-nlp-deploy-model).

When https://github.com/elastic/kibana/issues/144550 is complete and this action can be performed within Kibana, that should be mentioned here as well.

NOTE: The API mentions "assignment_state" whereas in this PR I'm using "allocation status", since that's what I see in the docs for the get trained model stats output. If it's not correct, we need to update for consistency.


szabosteve (approved):
Thank you for this change! LGTM.

PR #2246, opened 2022-11-08
[DOCS] Modifies regression API example

szabosteve:
## Overview

This PR sets the model memory limit to 50 MB in the regression API example to avoid running out of memory when users try the API example out.

### Preview

[Creating a regression model]() --available soon


valeriy42 (approved):
LGTM 👍

PR #2273, opened 2022-11-28
[DOCS] Add landing page for glossary

lcawl:
Since the glossary has only a single page, its initial landing page is quite empty:
image

We could create a simple landing page with links to the glossary's sections by using a `index-custom-title-page.html` page. For example:

image


gchaps:
Instead of a page of links, is it possible to put the glossary entries directly on the glossary landing page?  It would make for a better user experience.

PR #2282, opened 2022-12-07
[DOCS] Stop a single node

amyjtechwriter:
In line 19 of the [shut down node](https://github.com/elastic/elasticsearch/blame/master/docs/reference/upgrade/shut-down-node.asciidoc#L19) doc, it suggests `kill $(cat pid)` to stop a daemon process. Removed as not everyone is running Elasticsearch using a daemon. 

Fixes #88558

### DIFF
--- a/docs/en/install-upgrade/upgrading-elasticsearch.asciidoc
+++ b/docs/en/install-upgrade/upgrading-elasticsearch.asciidoc
@@ -109,7 +109,7 @@ jobs with large model states.
 . [[upgrade-node]] *Shut down a single node*.
 +
 --
-To shut down a single node depends on what is currently used to run {es}. For example, if using `systemd` or SysV `init` see the commands below.
+To shut down a single node depends on what is currently used to run {es}. For example, if using `systemd` or SysV `init` run the commands below.
 
 * If you are running {es} with `systemd`:
 +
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/upgrading-elasticsearch.asciidoc
```diff
@@ -109,6 +109,8 @@ jobs with large model states.
 . [[upgrade-node]] *Shut down a single node*.
 +
 --
+To shut down a single node depends on what is currently used to run {es}. For example, if using `systemd` or SysV `init` see the commands below.
```
kilfoyle, line 112:
```suggestion
To shut down a single node depends on what is currently used to run {es}. For example, if using `systemd` or SysV `init` run the commands below.
```
Our accessibility guidelines (which I can't find at the moment or I would share the link) recommend not to use `see`. Instead we typically use `check`, `refer to`, or other.

kilfoyle (approved):
LGTM! 🚗

docs/en/install-upgrade/upgrading-elasticsearch.asciidoc
amyjtechwriter: Ahhhh ok got it. I'll see if I can hunt them down. Thank you for catching that!

PR #2305, opened 2023-01-16
Adds docs about how anomaly scores are calculated

szabosteve:
## Overview

This PR adds a new page under the advanced concepts that helps users to understand how the anomaly scores are calculated, which factors contribute to an anomaly, and how renormalization works.

### Preview

[Anomaly score explanation](https://stack-docs_2305.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-ad-explain.html)


valeriy42:
Good work @szabosteve 🚀 I noticed you didn't add screenshots illustrating the individual impacts. Do you think they are redundant, given the final screenshot?

docs/en/stack/ml/anomaly-detection/ml-ad-explain.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+
+Three factors can affect the initial anomaly score of a record: 
+
+* single bucket impact,
+* multi-bucket impact,
+* anomaly characteristics impact.
+
+{anomaly-jobs-cap} split the time series data into time buckets. The data within 
+a bucket is aggregated using functions. Anomaly detection is happening on the 
+bucket values.
```
lcawl, line 27: > {anomaly-jobs-cap} split the time series data into time buckets....

IMO this paragraph fits at the beginning of this section, like this:

> {anomaly-jobs-cap} split the time series data into time buckets.... bucket values. Three factors can affect the initial anomaly score of a record: ...

That way it doesn't interrupt the discussion of the meaning of each factor

docs/en/stack/ml/anomaly-detection/ml-ad-explain.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+is the median value of the probability distribution for the bucket. This 
+probability leads to the single bucket impact. It usually dominates the initial 
+anomaly score of a short spike or dip.
+
+[discrete]
+[[multi-bucket-impact]]
+=== Multi-bucket impact
+
+The probabilities of the values in the current bucket and the preceding 11 
+buckets consist of the multi-bucket impact. The accumulated differences between
```
lcawl, line 46: > buckets consist of the multi-bucket impact....

Is "consist of" the right word here? Or should it be something like "contribute to" or "affect"?

docs/en/stack/ml/anomaly-detection/ml-ad-explain.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+
+
+[discrete]
+[[anomaly-characteristics-impact]]
+=== Anomaly characteristics impact
+
+The impact of the anomaly characteristics considers the different features of 
+the anomaly, such as its length and size. The total duration of the anomaly is 
+considered, and not a fixed interval as in the case of the multi-bucket impact 
+calculation. The length might be only one bucket or thirty, but even more.
```
lcawl, line 64: The "but even more" seems a bit unclear to me. Something like this is more straight-forward:
```suggestion
calculation. The length might be only one bucket or thirty (or more) buckets. 
```
It still make me wonder if there is an upper maximum number, but maybe that's too low-level of a detail.

docs/en/stack/ml/anomaly-detection/ml-ad-explain.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+
+[role="screenshot"]
+image::images/renormalization-score-reduction.jpg["Example of a record score reduction in {kib}"]
+
+
+[discrete]
+[[other-factors]]
+== Other factors for score reduction
+
+Two more factors may lead to a reduction of the initial score: high variance
```
lcawl, line 98: Not mandatory, but it feels like this needs "a" or "the":
```suggestion
Two more factors may lead to a reduction of the initial score: a high variance 
```

docs/en/stack/ml/anomaly-detection/ml-ad-explain.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+[role="screenshot"]
+image::images/renormalization-score-reduction.jpg["Example of a record score reduction in {kib}"]
+
+
+[discrete]
+[[other-factors]]
+== Other factors for score reduction
+
+Two more factors may lead to a reduction of the initial score: high variance 
+interval and incomplete bucket.
```
lcawl, line 99: Likewise, this feels like it needs "an" or an "-s":
```suggestion
interval and an incomplete bucket. 
```

docs/en/stack/ml/anomaly-detection/ml-ad-explain.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+
+[discrete]
+[[other-factors]]
+== Other factors for score reduction
+
+Two more factors may lead to a reduction of the initial score: high variance 
+interval and incomplete bucket. 
+
+{anomaly-detect-cap} is less reliable if the current bucket is part of a 
+seasonal pattern with high variability in data. For example, you may have server
```
lcawl, line 102: This also seems to need a "the" or "its" IMO:
```suggestion
seasonal pattern with high variability in its data. For example, you may have server 
```

lcawl (approved):
Some minor comments, but otherwise a really great read!

valeriy42 (approved):
LGTM. Good work @szabosteve!

PR #2311, opened 2023-01-26
Fixes GH links on Apache and Nginx OOTB job pages

szabosteve:
## Overview

This PR changes links on the Apache and Nginx pages that point to the GH repo where the spec files are stored. Until 8.1, it was the Kibana repo (version controlled). From 8.2, it's the Integrations repo (not version controlled). For this reason, the links from 8.2 to 8.6 are broken. This PR fixes this problem. 7.17 is also affected as it's an active (updated) branch.


lcawl (commented):
I think the other links in the table will also need to be updated too, right?

szabosteve:
Yup, you are right, @lcawl! I updated the PR accordingly.

lcawl (approved):
LGTM!

PR #2330, opened 2023-02-16
[DOCS] Adds info about available field statistics in AD and DFA wizards

szabosteve:
## Overview

This PR adds a TIP about the available field stats in the anomaly detection and the data frame analytics wizards.
The PR also removes a link that is not relevant anymore and fixes some line breaks.


qn895 (approved):

qn895:
LGTM 🎉 I wonder if the page `Predicting numerical values with regression` and `Finding outliers` should get a similar tip to indicate that these field stats are also available?

szabosteve:
@qn895 Yup, I failed to push those changes. Thanks for noticing it! Fixed in [0e8e9e2](https://github.com/elastic/stack-docs/pull/2330/commits/0e8e9e281d87614127227cf13e33f76bdc16f7b7).

PR #2335, opened 2023-02-21
Update NLP example to use semantic search

abdonpijpelink:
This PR updates the NLP text embedding example to use the semantic search capabilities released with 8.7. 

I've also updated a few instances of "vector search" with "semantic search", as that will make it easier for folks to find this example.


szabosteve (approved):
Thank you for modifying the example! 🎉 
LGTM!

davidkyle (approved):
LGTM

PR #2339, opened 2023-02-24
Update ML docs for NER support

benwtrent:
As of 8.4.0, Elasticsearch supports any IOB tagged NER entities.


szabosteve (approved):
Thank you, LGTM!

PR #2357, opened 2023-03-30
[DOCS] No snapshots are removed after a major version upgrade is complete.

amyjtechwriter:
Correction around snapshot behaviour. They are not removed after upgrade.


kilfoyle (approved):
LGTM! ⛵ 
Just one suggestion for the text.

PR #2369, opened 2023-04-21
[DOCS] Adds section title and anchor for model deployment in air-gaped env

szabosteve:
## Overview

This PR adds an empty section to the `Deploy trained models` page that will explain how to deploy trained models in an air-gaped environment. It also adds a section ID, so the section will have its own anchor (`#ml-nlp-deploy-model-air-gaped`) on the page.

Content will be added soon.


hendrikmuhs (approved):
LGTM

PR #2391, opened 2023-05-04
Fix ELSER download instructions

demjened:
Fix ELSER download instructions (via Enterprise Search) in `ml-nlp-elser.asciidoc`.

The end-to-end process requires two user actions:
- Download the model with "Deploy" button
- Start the model after it's been downloaded

### DIFF
--- a/docs/en/stack/ml/nlp/ml-nlp-elser.asciidoc
+++ b/docs/en/stack/ml/nlp/ml-nlp-elser.asciidoc
@@ -33,25 +33,35 @@ for semantic search or the trial period activated.
 
 
 [discrete]
-[[download-elser]]
-== Download ELSER
+[[download-deploy-elser]]
+== Download and deploy ELSER
 
-You can download ELSER either from **{ml-app}** > **Trained Models**, from 
-**{ents}** > **Indices**, or by using the Dev Console.
+You can download and deploy ELSER either from **{ml-app}** > **Trained Models**, 
+from **{ents}** > **Indices**, or by using the Dev Console.
 
 [discrete]
 [[trained-model]]
 === Using the Trained Models page
 
 1. In {kib}, navigate to **{ml-app}** > **Trained Models**. ELSER can be found 
-in the list of Trained Models.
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/nlp/ml-nlp-elser.asciidoc
```diff
@@ -65,11 +65,11 @@ You can also download and deploy ELSER to an {infer} pipeline directly from the
 2. Select the index from the list that has an {infer} pipeline in which you want 
 to use ELSER.
 3. Navigate to the **Pipelines** tab.
-4. Under **{ml-app} {infer-cap} Pipelines**, click the "Start single-threaded" 
-button to download and start the model with basic configuration or select the 
-"Fine-tune performance" option to navigate to the **Trained Models** page where 
-you can configure the model deployment.
-
+4. Under **{ml-app} {infer-cap} Pipelines**, click "Deploy" button to begin downloading
```
droberts195, line 68: Since there are two steps for Enterprise Search as well as the other methods it probably makes sense to structure this whole docs page similarly for all 3 methods.

At the moment we have high level "Download" and "Deploy" sections with the two steps split for the ML UI and dev console but Enterprise Search's two steps all in the first section. So it's probably better to either just have one high level section and put all the steps for each method in one list for that method. Or else split Enterprise Search too.

This also highlights that there is different terminology in different apps. ML is using "deploy" to mean "start" whereas Enterprise Search is using "deploy" to mean "download".

@szabosteve maybe you should take over this PR to do a broader refactoring.

szabosteve: Ack. Will do today.

droberts195 (approved):
LGTM

docs/en/stack/ml/nlp/ml-nlp-elser.asciidoc
demjened: Hey @szabosteve, it's all good 🙂  Yep, I'll forward-port.

PR #2402, opened 2023-05-19
[DOCS] Add frontmatter for migration

lcawl:
This PR updates source files that contain `:description:` attributes to use `:frontmatter-description:` instead. It also adds the other frontmatter necessary for migration to the new docs system.


szabosteve (approved):
Thank you for these! LGTM!

PR #2433, opened 2023-06-12
[DOCS] Adds more conceptual info to ELSER

szabosteve:
## Overview

This PR adds more conceptual information to the ELSER introduction about how the model works.

### Preview

[ELSER](https://stack-docs_2433.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-nlp-elser.html)

### DIFF
--- a/docs/en/stack/ml/nlp/ml-nlp-elser.asciidoc
+++ b/docs/en/stack/ml/nlp/ml-nlp-elser.asciidoc
@@ -20,9 +20,9 @@ meaning and user intent, rather than exact keyword matches.
 ELSER is an out-of-domain model which means it does not require fine-tuning on 
 your own data, making it adaptable for various use cases out of the box.
 
-ELSER expands the indexed and searched terms into larger collections of terms 
-that are learned to co-occur frequently within a diverse set of training data. 
-The terms that the text is expanded into by the model _are not_ synonyms for the 
+ELSER expands the indexed and searched passages into collections of terms that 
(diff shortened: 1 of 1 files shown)


docs/en/stack/ml/nlp/ml-nlp-elser.asciidoc
```diff
@@ -20,9 +20,13 @@ meaning and user intent, rather than exact keyword matches.
 ELSER is an out-of-domain model which means it does not require fine-tuning on 
 your own data, making it adaptable for various use cases out of the box.
 
-Refer to 
-https://www.elastic.co/blog/may-2023-launch-information-retrieval-elasticsearch-ai-model[this blog post], 
-to learn more about ELSER.
+ELSER expands the indexed and searched terms into larger collections of terms
```
qherreros, line 23: The expansion is not necessarily larger. Typically, a passage of 512 tokens would be expanded into 40/50 tokens. While doing hybrid retrieval, you could argue that you are using all previous terms with BM25 + all new terms with ELSER, so your total collection would be higher. But if you are using only ELSER, your collection of terms is actually smaller.

I would say:
ELSER expands the indexed and searched passages into collections of terms that are learned to co-occur frequently within a diverse set of training data.

szabosteve: Updated via [d4b80a0](https://github.com/elastic/stack-docs/pull/2433/commits/d4b80a0cd827814b8251942839899b7e6c6ff8c1).

PR #2435, opened 2023-06-14
Adds size parameter to the reindex commands in the NLP examples

szabosteve:
## Overview

This PR adds the `size` parameter to the reindex API calls in the NLP examples.


davidkyle (approved):
LGTM

PR #2310, opened 2023-01-25
Marks recommended models in the third party models list

szabosteve:
## Overview

Related issue: https://github.com/elastic/ml-team/issues/894
This PR highlights recommended models for every task type on the supported models' page. 

### Preview

[Compatible third party NLP models](https://stack-docs_2310.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-nlp-model-ref.html)

PR #2451, opened 2023-07-03
Update ml-nlp-model-ref.asciidoc

benwtrent:
Neither of the "recommended" models are actually answer extraction models. They are both sentence similarity models. Very different tasks.

Originally added via this change: https://github.com/elastic/stack-docs/pull/2310


szabosteve (approved):
Thanks for the fix, Ben! I'll forwardport the changes to `8.9` and `main`!

PR #2495, opened 2023-07-28
Remove + redirect breaking changes pages

jrodewig:
Updates the [Breaking changes](https://www.elastic.co/guide/en/elastic-stack/current/elastic-stack-breaking-changes.html) pages of the Install/Upgrade Guide to:

- Remove child pages that include includes from other books. These often break and are a pain to maintain.
- Replace links to the pages with links to the doc books.
- Add redirects for the child pages.


docs/en/install-upgrade/redirects.asciidoc
```diff
@@ -25,4 +25,40 @@ See the {observability-guide}/whats-new.html[Observability highlights] page for
...
+=== {kib} breaking changes
+
+This page no longer exists.
+See {kibana-ref}/release-notes.html[{kib} release notes] instead.
+
+[role="exclude",id="logstash-breaking-changes"]
+=== {ls} breaking changes
+
+This page no longer exists.
+See {logstash-ref}/breaking-8.0.html[{ls} 8.0 breaking changes] instead.
```
karenzone, line 64:
```suggestion
See {logstash-ref}/breaking-changes.html[{ls} breaking changes] instead.
```

karenzone: What about linking to the Logstash Breaking changes page?  The additional context helps users understand that we don't have breaking changes for every minor.  If we did have breaking changes for an 8.x release, a user could easily find them. 

This page would be evergreen and not require updating if/when we move to 9.x.

jrodewig: Great suggestion! Done in [`277d0d8` (#2495)](https://github.com/elastic/stack-docs/pull/2495/commits/277d0d8262636db03bc28c133e78133768a4e7a9)

bmorelli25 (approved):
😎

PR #2513, opened 2023-08-09
Fix link to Enterprise Search release notes

chriscressman:
The current link points to a specific version of the release notes rather than the release notes landing page. Point to the landing page instead.

This makes it easier to bump the Elastic docs version, since it removes a dependency on a specific Enterprise Search page existing first.

It's also consistent with the other release notes links on this page.

Rel: https://github.com/elastic/stack-docs/pull/2495


jrodewig (approved):
Good catch. Thanks for fixing this.

I added related tags to backport this to 8.9.

PR #2531, opened 2023-08-24
[DOCS] Add default ports for Elastic Stack components

jrodewig:
Documents the default ports for major Elastic Stack components.

### Progress, SIMPLE Perfection
The aim here is to be accurate but not exhaustive. Good > Perfect.

I plan to add ports for ECE and ECK in a later PR.
I'll also add links to settings in a later PR.

### DIFF
--- a/docs/en/install-upgrade/installing-stack.asciidoc
+++ b/docs/en/install-upgrade/installing-stack.asciidoc
@@ -9,8 +9,44 @@ Kibana {version}, and Logstash {version}.
 If you're upgrading an existing installation, see <<upgrading-elastic-stack, Upgrading the Elastic Stack>> for information about how to ensure compatibility with {version}.
 
 [discrete]
+[[network-requirements]]
+=== Network requirements
+
+// https://support.elastic.dev/knowledge/view/CfT70cu9
+
+The following ports need to be open for components of the Elastic Stack.
+
+[cols="1,1,1"]
+|===
+|Default ports |Access type| Component
+
+|5601
+|HTTP
+|{kib}
+
+|8200
+|HTTP
+|APM Server
+
+|9200-9300
+|HTTP
+|{es} REST API
+
+|9300-9400
+|TCP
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/installing-stack.asciidoc
```diff
@@ -8,9 +8,45 @@ Kibana {version}, and Logstash {version}.
...
+| 5044
+| {agent} → {ls} +
+{beats} → {ls}
+
+|5601
+|{kib} +
+{agent} → {fleet} +
+{fleet-server} → {fleet}
+
+|8200
```
nimarezainia, line 30: I believe this is 8220

nimarezainia (approved):

docs/en/install-upgrade/installing-stack.asciidoc
jrodewig: :doh: Thanks so much for catching this.

liladler:
Thanks @jrodewig that's fantastic! It's great to see feedback from customers and field gets included in the docs :) 
Also great it refers to the ingest docs for reference architectures.
A few comments from my side:
 - The reference architecture focuses on ingest so only shows breakdown of ingest components. Would be great to have a reference architecture that presents Elasticsearch components, such as ML nodes, Hot, Cold, Frozen buckets etc.
 - The page references APM/fleet server. In the Cloud console we still call it integration server. might worth a clarification that it refers to the same component?
 - Enterprise search ports aren't referenced
 - Worth mentioning that each integration has its own ports and dependencies, so this should be verified ahead of installation.
 - We're referencing cloud, but cloud ports are different (uses TLS).

Apart from that looks good. Thanks again for this!

jrodewig:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.10|[](https://github.com/elastic/stack-docs/pull/2533)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

PR #2548, opened 2023-09-26
Add list of text similarity models to compatible NLP models

davidkyle:
Text similarity must use a cross encoder model which has caused some confusion

For https://github.com/elastic/elasticsearch/issues/99211


szabosteve (approved):
Thank you for this addition! LGTM.

PR #2550, opened 2023-09-27
Changes to Beats and Logstash minimal version

joeafari:
Support matrix for product compatibility indicates that only Beats and Logstash version 7.17.x are compatible with 8.x
See https://www.elastic.co/fr/support/matrix#matrix_compatibility

Aligning this to that as the matrix is the source of truth


joeafari:
Had a discussion about this: https://elastic.slack.com/archives/C0461HZ51NG/p1695808818909129
It is compatible (it should work, we don't block this) but not supported (we've tested this).

PR #2555, opened 2023-10-05
[DOCS] Highlights how to switch to ELSER v2 from v1

szabosteve:
## Overview

This PR adds a subsection to the ELSER v2 section about how to upgrade from v1 to v2. It also adds a link that points to the tutorial which shows how to create an index pipeline and how to reindex using that pipeline.

### Preview

[ESLER v2]()


davidkyle (approved):
LGTM

PR #2567, opened 2023-10-17
[DOCS] Documents that models with one allocation might have downtime

szabosteve:
## Overview

This PR adds an `IMPORTANT` block to the trained model deployment section that explains that users will experience downtime in the service of their trained model if the model has only one allocation deployed. 

### Preview

[Deploy the model to your cluster](https://stack-docs_2567.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-nlp-deploy-model.html#ml-nlp-deploy-model)


droberts195:
> serverless projects only

No, it affects all types of installation - on-premise, classic Cloud and serverless. If a model is only running on one node then it won't be able to service requests when that node gets restarted for any reason.

wwang500 (approved):
LGTM. 👍

droberts195 (approved):
LGTM

PR #2576, opened 2023-10-19
Improve rolling upgrade order docs

DaveCTurner:
The first point in the list (upgrade the master-ineligible nodes) is
redundant and sort of contradicts the second one (upgrade the data nodes
tier-by-tier). This commit drops it.

The second point had some redundant sentences in a somewhat awkward
order. This commit cleans that up.

There's no mention of nodes that are neither data nodes nor
master-eligible nodes. In fact it doesn't really matter when these nodes
are upgraded, but this commit specifies to upgrade them before the
master-eligible nodes for the sake of completeness.

Finally, there were some over-long lines so this commit reflows the
source.


DaveCTurner:
Not sure about the labelling policy in this repo but my intention is for this to go into the current live docs (8.10) and future versions (8.11 and beyond).

abdonpijpelink (approved):
Looks great. Thanks @DaveCTurner !

PR #2602, opened 2023-12-01
Add Elastic Agent to stack installation instructions

dedemorton:
When reviewing our new landing page design, I noticed that Elastic Agent is completely missing from this list. 

I'm not sure why we are hard coding the product names, but decided to leave this as-is in case there is a reason.

@kilfoyle I hope I am not overstepping, but opening a PR was faster than creating an issue. :-)

### DIFF
--- a/docs/en/install-upgrade/installing-stack.asciidoc
+++ b/docs/en/install-upgrade/installing-stack.asciidoc
@@ -61,7 +61,7 @@ Install the Elastic Stack products you want to use in the following order:
 . Elasticsearch ({ref}/install-elasticsearch.html[install instructions])
 . Kibana ({kibana-ref}/install.html[install])
 . Logstash ({logstash-ref}/installing-logstash.html[install])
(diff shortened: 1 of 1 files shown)


kilfoyle (approved):
Thanks @dedemorton! Definitely we should have agent here.

Kind of related, I have an issue somewhere to add agent and APM to the [upgrade on -prem](https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elastic-stack-on-prem.html) page. I'll add them both in, but if you have thoughts on where APM fits in the upgrade order, please let me know.

dedemorton:
@kilfoyle Good question. I would double-check with the APM team (maybe @colleenmcginnis knows?) becaseu there are a couple ways to run APM. The [APM upgrade docs](https://www.elastic.co/guide/en/apm/guide/current/upgrading-to-8.x.html) have a lot of caveats, and I'm not sure there is a simple answer here. 

BTW, I've added the 8.11 backport label but I'm not sure what other versions we want to backport this to. Let me know!

PR #2612, opened 2023-12-07
Edit github workflow for documentation previews

lcawl:
Relates to https://github.com/elastic/docs/pull/2848

Since this repo is used to generate multiple books, I set the `preview-path` to the top-level landing page instead of a particular book.

It seems that this change will need to be backported all the way to 8.4.

### DIFF
--- a/.github/workflows/docs-preview-links.yml
+++ b/.github/workflows/docs-preview-links.yml
@@ -1,26 +1,19 @@
-name: Docs Preview Links
+name: docs-preview
 
 on:
   pull_request_target:
     types: [opened]
-    paths:
-      - '**.asciidoc'
+
+permissions:
+  pull-requests: write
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-preview-links.yml
```diff
@@ -1,26 +1,19 @@
-name: Docs Preview Links
+name: docs-preview
 
 on:
   pull_request_target:
     types: [opened]
-    paths:
-      - '**.asciidoc'
```
nkammah, line 7: You can consider keeping the above condition if you don't want the action to post a comment on non doc changes.

lcawl: Good point, I've re-added it via c9fe55a91e15cb2c389ea179da4584d72eb92f2c

PR #2624, opened 2024-01-10
Adds reference to E5 model in ELSER docs

szabosteve:
## Overview

This PR adds a note to the ELSER docs that ELSER is an English language model, so if the user would like to perform semantic search on non-English documents, use E5 instead.

### Preview

[ELSER](https://stack-docs_2624.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-nlp-elser.html)


davidkyle (approved):
LGTM

PR #2660, opened 2024-01-29
Suggest chunking for large ELSER fields

seanstory:


seanstory:
I don't seem to have permission to modify labels. This should probably get backported. CC @leemthompo I'm hoping you can shepherd this? 🙏 (couldn't explicitly add you as a reviewer, either)

szabosteve (approved):
Thank you for this addition! LGTM!

PR #2706, opened 2024-05-03
Update installing-stack-demo-secure.asciidoc

eric-cobb:
File path didn't match what we listed in Step 3.1.vii


kilfoyle (approved):
LGTM! 🚀

PR #2744, opened 2024-07-05
Adds important callout about the possibility of multiple ELSER deployment

szabosteve:
## Overview

It is possible to deploy ELSER multiple times by using different deployment IDs. The recommended behavior is to have different deployments with different purposes. This PR adds an `important` callout to the ELSER page to highlight these pieces of information.


kilfoyle (approved):
LGTM! 🦒

PR #2948, opened 2025-02-07
Add upgrade note about release date constraint

DaveCTurner:
We don't support (and will actively block) upgrades to
chronologically-older versions. This commit adds a note in the docs
about this.

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -10,14 +10,15 @@ These steps vary based on your current version:
 endif::[]
 
 [IMPORTANT]
+.Out-of-order releases
 ====
-Elastic maintains several minor versions of the Elastic Stack at once. This means
-releases do not always happen in order of their version numbers. You can only
-upgrade to {version} if the version you are currently running meets both these
-conditions:
+Elastic maintains several minor versions of the Elastic Stack at once. This
+means releases do not always happen in order of their version numbers. You can
+only upgrade to {version} if the version you are currently running meets both
+of these conditions:
 
-* Has an older version number than {version}
-* Has an earlier release date than {version}
+* Has an older version number than {version}.
+* Has an earlier release date than {version}.
 
 If you are currently running a version with an older version number but a later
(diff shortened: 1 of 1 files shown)


DaveCTurner:
Not sure how to label these things, sorry, please could you apply appropriate labels. This one doesn't need backporting, this constraint only applies in practice from 8.19 onwards.

DaveCTurner:
Correction, sorry, this will also apply to upgrades to 8.18, so yes needs backporting to the 8.18 branch.

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -9,6 +9,17 @@ These steps vary based on your current version:
...
+[NOTE]
+====
+Elastic maintains several minor versions of the Elastic Stack at once, so
+releases do not always happen in order of their version numbers. You can only
+upgrade to {version} if the version you are currently running has an older
+version number than {version} _and_ has an earlier release date than {version}.
+
+If you are currently running a version with an older version number but a later
+release date than {version}, wait for a newer release before upgrading.
+====
```
leemthompo, line 21:
```suggestion
[IMPORTANT]
====
Elastic maintains several minor versions of the Elastic Stack at once. This means
releases do not always happen in order of their version numbers. You can only
upgrade to {version} if the version you are currently running meets both these
conditions:

* Has an older version number than {version}
* Has an earlier release date than {version}

If you are currently running a version with an older version number but a later
release date than {version}, wait for a newer release before upgrading.
====
```

leemthompo: Couple minor optional tweaks for readability + stronger admonition type

leemthompo: Maybe "Upgrades to chronologically-older versions will be blocked" is helpful info too?

DaveCTurner: ++ thanks, updated & tweaked a little further

leemthompo (approved):
LGTM

PR #2949, opened 2025-02-07
[Backport][8.18] Add upgrade note about release date constraint (#2948)

leemthompo:
"Backports" https://github.com/elastic/stack-docs/pull/2948 to 8.18 because mergify wasn't configured to backport from any branches other than `main`

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -9,6 +9,19 @@ These steps vary based on your current version:
 * <<prepare-to-upgrade,Upgrade from 7.x>>
 endif::[]
 
+[IMPORTANT]
+.Out-of-order releases
+====
+Elastic maintains several minor versions of the Elastic Stack at once. This
+means releases do not always happen in order of their version numbers. You can
+only upgrade to {version} if the version you are currently running meets both
+of these conditions:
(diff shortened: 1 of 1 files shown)


georgewallace (approved):

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -9,6 +9,19 @@ These steps vary based on your current version:
...
+====
+Elastic maintains several minor versions of the Elastic Stack at once. This
+means releases do not always happen in order of their version numbers. You can
+only upgrade to {version} if the version you are currently running meets both
+of these conditions:
+* Has an older version number than {version}.
+* Has an earlier release date than {version}.
+If you are currently running a version with an older version number but a later
+release date than {version}, wait for a newer release before upgrading.
+====
```
leemthompo, line 23:
```suggestion
[IMPORTANT]
.Out-of-order releases
====
Elastic maintains several minor versions of the Elastic Stack at once. This
means releases do not always happen in order of their version numbers. You can
only upgrade to {version} if the version you are currently running meets both
of these conditions:

* Has an older version number than {version}.
* Has an earlier release date than {version}.

If you are currently running a version with an older version number but a later
release date than {version}, wait for a newer release before upgrading.
====
```

PR #3013, opened 2025-04-03
Fixes redundancy in the Transforms migration section

kosabogi:
This PR removes redundant sentences from the Transforms destination indices section.

### DIFF
--- a/docs/en/install-upgrade/upgrading-stack.asciidoc
+++ b/docs/en/install-upgrade/upgrading-stack.asciidoc
@@ -526,7 +526,7 @@ You can use the https://www.elastic.co/docs/api/doc/elasticsearch/operation/oper
 GET _cat/indices/.transform-destination-example?v&h=index,store.size
 ----
 
-If an index size is greater than 10 GB, it is recommended to use the Reindex API. Reindexing consists of the following steps:
+If an index size is greater than 10 GB, we recommend using the Reindex API. Reindexing consists of the following steps:
 
 . Set the original index to read-only.
 +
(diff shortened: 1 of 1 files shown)


docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -519,17 +519,13 @@ When {kib} Upgrade Assistant reindexes the documents, {kib} will put a write blo
 
 If an index is less than 10GB of size, we recommend using {kib}'s Upgrade Assistant to automatically migrate the index.
```
jmikell821, line 520:
```suggestion
If an index size is less than 10 GB, we recommend using {kib}'s Upgrade Assistant to automatically migrate the index.
```

docs/en/install-upgrade/upgrading-stack.asciidoc
```diff
@@ -519,17 +519,13 @@ When {kib} Upgrade Assistant reindexes the documents, {kib} will put a write blo
...
 You can use the https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cat-indices-1[Get index information API] to obtain the size of an index:
 
 [source,console]
 ----
 GET _cat/indices/.transform-destination-example?v&h=index,store.size
 ----
 
-The reindexing can be initiated in the {kib} Upgrade Assistant.
-
 If an index size is greater than 10 GB, it is recommended to use the Reindex API. Reindexing consists of the following steps:
```
jmikell821, line 529:
```suggestion
If an index size is greater than 10 GB, we recommend using the Reindex API. Reindexing consists of the following steps:
```

PR #3014, opened 2025-04-04
Potential typo in ml-nlp-elser.asciidoc

seanhandley:
Seems like a typo but happy to be corrected


szabosteve (dismissed):
Thank you!

szabosteve (commented):
Thank you for being helpful!
The docs are correct. ELSER V1 is an experimental feature and remains experimental. V2 is GA.

PR #3029, opened 2025-11-26
Fixes broken links on "Security anomaly detection configurations" page

nastasha-solomon:
Fixes broken links to ML jobs and datafeeds on the "Security anomaly detection configurations" page.

Preview


nastasha-solomon:
Don't need

PR #3032, opened 2025-12-23
[Docs] Increase vm.max_map_count suggested value

shainaraskas:
We want to recommend a higher `vm.max_map_count` for Elasticsearch environments - additional context in issue

## What issues does this PR fix?

Part of https://github.com/elastic/docs-content-internal/issues/592


shainaraskas:
this page does not exist in previous versions and should be redirected instead. opened an issue for that 

https://github.com/elastic/docs-content-internal/issues/615

